PR #24, opened 2016-08-13
Align class name with file name to fix error on case-sensitive filesy…

jotomo:
…stems/OSs.

@MilosKozak On a related note: just importing the project into Android Studio on Linux, several files under the _.idea_ directory are changed. I was thinking of using the [_.gitignore_ Intellij plugin](https://github.com/hsz/idea-gitignore) to generate a _.gitignore_ file that excludes IntelliJ and Eclipse files more thoroughly. Sounds good?


MilosKozak:
ok. one time i did the mistake and changed cases in windows which is not case sensitive FS and all troubles are from there. i did a test clone and it's cloned again in uppercase.  so your pr is correct

PR #25, opened 2016-08-14
Update .gitignore, remove newly ignored files from git.

jotomo:
Update .gitignore, remove newly ignored files from git.

Existing files in working directories are not touched (git rm --cached $ignored_file).

This should make it easier for contributors to collaborate without git getting in the way.


MilosKozak:
did you use android studio or intellij? looks like it offers to remove potentially important project files

jotomo:
I'm using Android Studio 2.1.2. The files in the diff are only removed from git, your local files won't be touched. Though it can't hurt to make a copy of the project folder beforehand, in the event I overlooked something.

My motivation for this was that right after opening the project in Android Studio, a lot of files were changed by AS and thus flagged by git as changed.
I just tested opening a fresh checkout where the ignored files are not there. The import works, the  remaining file that is changed is _.idea/modules.xml_ ...
I tried to use the _.ignore_ plugin recommended settings and not exclude too much. My usual (brute-force, but simple and working) approach is to exclude the _.idea_ directory entirely and have AS generate the folder based on build.xml/pom.xml since that file is the more relevant.

jotomo:
*sigh* PR still has issues. I have a usable workaround to use locally, so I'll close this, so more important things can be attended to.

PR #37, opened 2016-10-26
Update SmsCommunicatorPlugin.java

warstar2187:
"newSms = new Sms(receivedSms.phoneNumber, reply, new Date());" is missing on some sms command 
Added it on this update.


MilosKozak:
👍 i forget these lines during last refactoring.
Thanks

PR #40, opened 2016-10-29
update bg translation

slavidex:


MilosKozak:
add `formatted="false"` where needed

slavidex:
OK, but I think it's a mistake to translate SMS commands. They may  not work in Bulgarian

PR #41, opened 2016-10-31
Add percent to result to allow manual adjust

ecorman:
My pump can only be adjusted in percent values, but only 0-10-20...250 values are allowed, not sure how to do this in good way.


ecorman:
Sorry I use virtualPump instead of configbuilder... I propose you another way, in any case I'll appreciate if you can give me feedback if it's correct or not. I'm learning sorry

PR #42, opened 2016-10-31
Show current total

AdrianLxM:
The total of the calculation in the wizard was showing the value from the previous calculation when you enable/disable to include Bolus-IOB or Basal-IOB into the calculation.

screenshot_2016-10-31-12-30-23

The button to send to the pump showed the current amount of units.


MilosKozak:
thx

PR #55, opened 2016-11-09
Upload only

AdrianLxM:
Tested:

* Adding a treatment via NS works on "upload only" disabled.
* Adding a treatment via NS will be blocked on "upload only" enabled.
* Modifying a treatment will be blocked on "upload only".
* SGV still come through even if "upload only" enabled.
* Button in Treatments-Fragment will just show a toast if "upload only" is enabled.
* Button in Treatments-Fragment works (and downloads old treatments) if "upload only" is disabled.


AdrianLxM:
Added that profiles will still come through if NS-Profile plugin is activated.

AdrianLxM:
Closing in favour of https://github.com/MilosKozak/AndroidAPS/pull/59 that resolves the merge conflicts.

PR #61, opened 2016-11-10
Fill action

AdrianLxM:


MilosKozak:
keeping same IDs across multiple dialogs is fine?

AdrianLxM:
@MilosKozak 
> keeping same IDs across multiple dialogs is fine?

Yes. Only one of the dialogs can be active at the time.

PR #62, opened 2016-11-11
Show DanaR as part of profile name

AdrianLxM:
Now that the profile name is shown in the overview, prefixing the profile names returned from the DanaR profile plugin seems to make sense to me:

screenshot_2016-11-11-01-36-05


MilosKozak:
yeah, to have there only "1" would be confusing

PR #66, opened 2016-11-12
Deltas per5 minutes

AdrianLxM:


MilosKozak:
is this now equal to oref0 dev?

AdrianLxM:
No, it is not equal oref0 dev but closer.
It just fixes the issue that the deltas were not "per 5 minutes". (Less cancelling basal rates just to set them again after 5 minutes, as "expected delta < delta" was falsely detected).

The new logic from oref0 dev would be a much bigger change and would need further testing. Also the average-delta got replaced in dev by two averages `short_avgdelta` and `long_avgdelta` representing 15 and 30 minutes. `determine_basal` then does the the reasoning and  future prediction based on both.
Depending on the gap in data, the `short_delta` may very well be `undefined`(or `0` in that case). Then for low prediction, the 30 minute delta would be used.

I think we can only switch to the full current dev logic if we also use the newer determine basal.

PR #68, opened 2016-11-15
Circadian percentage profile

AdrianLxM:


MilosKozak:
did you test compile release?

AdrianLxM:
Thanks for catching this. I only had tested it on the debug build.
Now it is also tested/working on the release build.

PR #81, opened 2016-11-29
Wear support

AdrianLxM:
Adds wear-support.

When compiling "Release" and a product flavour ending in "wear", the wear settings app and the watchfaces will get packed into the signed apk and pushed to paired android wear watch automatically on installation of the apk.

When compiling "Debug" and a product flavour ending in "wear", the wear-apk needs to be pushed to the watch separately (via the bluetooth-debugging interface e.g.).

When selecting a product flavour ending in "nowear", the wear apk will not be compiled and the mobile app will not show the possibility to enable the wear plugin.


MilosKozak:
I don't have wear so I'll rely on you ;)
merging

PR #82, opened 2016-12-01
Wear nullprofile fix

AdrianLxM:


MilosKozak:
make the toast message translatable pls. or pickup some already existing

AdrianLxM:
Thanks for the catch. Updated it.

MilosKozak:
i'd carefull with `this`. the fragment may not exist all the time. i hit this issue many times so i better use `MainApp.sResources.getString(....)` which always works

MilosKozak:
the same with getString() .... i better use `MainApp.sResources.getString(...)`

AdrianLxM:
Even if the fragment does not exist (or the GUI at all): `this` is an instance of a subclass of `Service` which is a context(Wrapper) by itself.

PR #83, opened 2016-12-03
Koreandanar

MilosKozak:


app/src/main/java/info/nightscout/androidaps/Services/DataService.java
```diff
@@ -96,11 +96,10 @@ protected void onHandleIntent(final Intent intent) {
                 ObjectivesPlugin.bgIsAvailableInNS = true;
                 ObjectivesPlugin.saveProgress();
             } else if (isNSProfile && Intents.ACTION_NEW_PROFILE.equals(action)){
-                // always handle Profili if NSProfile is enabled without looking at nsUploadOnly
+                // always handle Profile if NSProfile is enabled without looking at nsUploadOnly
                 handleNewDataFromNSClient(intent);
             } else if (!nsUploadOnly &&
-                    (Intents.ACTION_NEW_PROFILE.equals(action) ||
```
AdrianLxM: @MilosKozak, was easy to find more for me... much more code from you! ;)
The new logic means, if the DanaR profile is used, a new profile from NightScout will not be written to the pump?
Not a bad idea.

PR #84, opened 2016-12-05
basic notification functionality

MilosKozak:


MilosKozak:
image
@AdrianLxM any comments? do we need more functionality for the beginning?

MilosKozak:
expecting it'd be possible to show it/acknoledge on wear too if you add some small piece of code ..

app/src/main/java/info/nightscout/androidaps/plugins/Overview/NotificationStore.java
```diff
@@ -0,0 +1,60 @@
...
+                return;
+            }
+        }
+    }
+
+    public void removeExpired() {
+        for (int i = 0; i < store.size(); i++) {
+            Notification n = get(i);
+            if (n.validTo.getTime() != 0 && n.validTo.getTime() < new Date().getTime()) {
+                store.remove(i);
```
AdrianLxM: Won't this call of `remove` make the size one smaller and give the "next" element the index of curren `i`?
This might mean the emelent after a removed one is skipped?

AdrianLxM: `i--;` or similar is what I'm thinking of.

MilosKozak: right

PR #85, opened 2016-12-05
limit readings to values above 38

AdrianLxM:
Not tested yet.
@MilosKozak, I wonder if we need  the restrictions in the db queries as well?


MilosKozak:
better yes. did  you test it?

AdrianLxM:
Yes. It took me some while until I could force some low values (testing on life system).
Low values will now be ignored and AndroidAPS goes into "stale data" mode (no delta shown, number in overview struck through)

PR #88, opened 2016-12-10
Korean rm640g

AdrianLxM:
Fully remove mm640g code stubs.
Also solves the issue with AndroidAPS complaining that `mm640g` has translations but does not exist in the general strings.xml.


MilosKozak:
thanks

PR #110, opened 2017-01-03
use android.text.format.DateUtils

MilosKozak:


AdrianLxM:
I didn't test this on my phon, but for testing I used the same method to format the time:
https://github.com/MilosKozak/AndroidAPS/compare/master...AdrianLxM:timeformats

... and it seemed to work fine on the tested phones :)

It got rid of the Android bug that on some phones that are default in 24h format the settings won't get read properly and it showed am/pm format after restart until 24h was actively set again.

MilosKozak:
hmm don't know how to make it show months as numbers correctly by selected locale. adding DateUtils.FORMAT_NUMERIC_DATE displays date in english notation

PR #125, opened 2017-01-13
Extended restricted by MAXBOLUS DanaR-plugin

AdrianLxM:
KreanDanaR seems to use the same class from the other plugin.


AdrianLxM:
sorry, wrong branch, closing.

PR #126, opened 2017-01-13
 Extended restricted by MAXBOLUS DanaR-plugin

AdrianLxM:
KreanDanaR seems to use the same class from the other plugin.

(same as https://github.com/MilosKozak/AndroidAPS/pull/125 but to dev branch)


MilosKozak:
thanks

PR #131, opened 2017-01-17
Tempt pill

AdrianLxM:
... also refactored out the default min/max/target levels as they are now used in OpenapsMA, OpenapsAMA and Overview.


app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
```diff
@@ -588,6 +607,36 @@ public boolean onLongClick(View view) {
...
+                return true;
+            }
+        });
+        tempTargetView.setLongClickable(true);
+
+
+
+
+
+        if (profile == null || !pump.isInitialized()) {
```
MilosKozak: this is moved to the bottom of func

app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
```diff
@@ -588,6 +607,36 @@ public boolean onLongClick(View view) {
...
+            // disable all treatment buttons because we are not able to check constraints without profile
+            wizardButton.setVisibility(View.INVISIBLE);
+            treatmentButton.setVisibility(View.INVISIBLE);
+            return;
+        } else {
+            wizardButton.setVisibility(View.VISIBLE);
+            treatmentButton.setVisibility(View.VISIBLE);
+        }
+
```
MilosKozak: remove up to here

PR #137, opened 2017-01-21
Update Spanish Translation

ecorman:


app/src/main/res/values-es/strings.xml
```diff
@@ -340,5 +340,51 @@
...
+    <string name="activity">Actividad</string>
+    <string name="alert_dialog_storage_permission_text">Por favo reinicia el teléfono o AndroidAPS desde ajustes de sistema, sino AndroidAPS no guardara registros (importantes para trazar y verificar que el algoritmo funciona correctamente)</string>
+    <string name="array_of_elements">Matriz de %d elementos. Valor actual:</string>
+    <string name="basal_rate">Ratio Basal:</string>
+    <string name="basalvaluebelowminimum">Valor basal menor del mínimo. Perfil no fijado.</string>
+    <string name="base_profile_label">Perfil Base</string>
+    <string name="button1">Botón 1</string>
+    <string name="button2">Botón 2</string>
+    <string name="button3">Botón 3</string>
+    <string name="careportal_temptarget">Perfil Temporal</string>
```
AdrianLxM: En las otras traducciónes se traduce "target" con "objectivo".
Creo que se puede confundir con "perfil" - "profile" en inglés - que tambien tenemos.

AdrianLxM: (In other places "target" is translated with "objectivo. I think "perfil" - "profile" in english - can be mixed up with the "profile" we also have)

ecorman: Yes you're right, thanks for your review. I created a new PR to correct this issue, not sure if it's done in correct way. Github is not easy for me.

PR #143, opened 2017-01-24
Settings for deltas

AdrianLxM:
Includes the suggestion by Scott to use the *short average delta* as delta on unfiltered sources like xDrip/G4 when values get noisy.

Also gives the option to use the *short average delta* as delta when the delta cannot be computed - similar to the corrected "OpenAPS MA" behaviour.


MilosKozak:
common user will not know what to choose here. this makes it complicated. i think we have following options:
1) select one default option and hardcode it to config (and let andvanced users change it there)
2) choose the best option and use it
3) create something like engeneering mode

not sure what's better :(

LadyViktoria:
how about adding this option to constants :-)

PR #144, opened 2017-01-25
remove round-basal.js

AdrianLxM:


MilosKozak:
https://github.com/MilosKozak/AndroidAPS/pull/144/files#diff-dfb8e2a47d2b9d1ab398c64e51707846L154
this is probably not needed anymore too

AdrianLxM:
Yes, I also think this can be removed as it is in `basal-set-temp.js` now. I'll add it to this PR later then we have the V8 call changes grouped together.

PR #147, opened 2017-01-28
fix bgi in autosens

AdrianLxM:
bgi's most of the time where 0 or got truncated to full integers.

Example from tonight: autosens result factor from 1.0 before and 1.09 after this.


MilosKozak:
thanks!

PR #152, opened 2017-02-07
Wear bolusprogress

AdrianLxM:
This will show a card on android wear watches while a bolus is being delivered with the current percentage and status on it:
screen screen2 
screen3 screen4

The card-notification has an action (left swipe) added to it, that will cancel/stop the current bolus.

Internally I added a new status event `EventBolusRequested` that will be sent out as soon as the user requested a bolus as progress or bolus-start is only reported after a successful connection. This way it gives the user the possibility to cancel the bolus while the pump is still connecting.

There is a new setting on the watch where the user can specify if the watch should vibrate during bolus delivery or not.


AdrianLxM:
closing this - the whole controll from the watchalready is much further along.

PR #158, opened 2017-02-10
use validation lib in cpp

LadyViktoria:


app/src/main/res/layout/circadianpercentageprofile_fragment.xml
```diff
@@ -44,23 +47,34 @@
                         android:src="@mipmap/ic_percentage"
                         android:textAppearance="?android:attr/textAppearanceMedium"/>
 
-                    <EditText
+                    <com.andreabaccega.widget.FormEditText
+                        validate:testType="numericRange"
+                        validate:minNumber="50"
+                        validate:maxNumber="200"
+                        android:digits="0123456789.,"
```
AdrianLxM: Before we had numbers without "." or ",".
Does this change anything?

LadyViktoria: yes it allows to enter , . and get an error.
without this maybe a user want to enter 20.0 % -> then he press . and nothing happen and 200% is there...

LadyViktoria: , and . are always present on number keyboard. but sometimes not clickable. to make them clickable it makes possible to prevent false insput

PR #166, opened 2017-02-11
cleanup branch

LadyViktoria:


MilosKozak:
did you try new clone and then build?

LadyViktoria:
deleted from hdd -> clone -> import as new project into studio -> build ok
also signed build is ok.
but also without reimporting. studio want to add/reload vcs root or something...
running studio 2.2.3

PR #167, opened 2017-02-11
fix deprecated methods

LadyViktoria:
try to fix deprecated methods
do not merge it 
its just here to see if travis is fine with the cahnges
a lot things to do


app/src/main/java/info/nightscout/androidaps/tabs/SlidingTabLayout.java
```diff
@@ -151,7 +151,7 @@ public void setViewPager(ViewPager viewPager) {
 
         mViewPager = viewPager;
         if (viewPager != null) {
-            viewPager.setOnPageChangeListener(new InternalViewPagerListener());
+            viewPager.addOnPageChangeListener(new InternalViewPagerListener());
```
AdrianLxM: On `EventRefreshGui` events, `setupTabs()` is called again without a new creation of the MainAcivity and therefore without a new instantiation of the viewPager (just finding it again by ID).
Doesn't that mean that switching from setting the listener (meaning: there can always be one) to adding the listener (to a list of listeners), it will get added over and over again during the life time of the MainActivity (and slow down performance)?

To emulate the same behaviour, that there is only one Listener, 
`viewPager.clearOnPageChangeListeners()` could be called prior to adding a new one.

build.gradle
```diff
@@ -16,6 +16,9 @@ allprojects {
     repositories {
         jcenter()
     }
+    tasks.withType(JavaCompile) {
+        options.compilerArgs << "-Xlint:deprecation" << "-Xlint:unchecked"
```
AdrianLxM: Why disable unchecked warnings globally (and for all future mistakes)?

LadyViktoria: around 200 cases...
https://travis-ci.org/MilosKozak/AndroidAPS/jobs/200569899
most of the cases are Date cases -> we should use calendar
and the other most case is getColor

LadyViktoria: now 21 build warnings.....
most of them with date... do not know how to handle this.
because there is no info about the date formatting

app/src/main/java/info/nightscout/androidaps/tabs/SlidingTabLayout.java
MilosKozak: agree. change it back. this must be tested separately

build.gradle
MilosKozak: for Date we should add new function to DateUtil and use it

MilosKozak: separate different things to more PR's to be able test it easier

PR #174, opened 2017-02-12
use validation lib in cpp - layout fixes & longer snackbar duration

LadyViktoria:


app/src/main/res/layout/circadianpercentageprofile_fragment.xml
```diff
@@ -3,6 +3,10 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
+    android:id="@+id/circadianpercentageprofile_linearlayout"
```
AdrianLxM: Hi Viktoria, it would be nice if you could rename this ID to framelayout. Thanks.

PR #177, opened 2017-02-13
Deprecation - remove wakelook and set it in bolusprogress dialog

LadyViktoria:


AdrianLxM:
Did you test this with calling the methods with the phone's screen turned off (on the wear bolus branch)?
A wakelock is far more powerfull than the flag to keep the screen on while a dialog is already visible.

The user can call at least `deliverTreatment()` without the gui being turned on from the watch or via the SMSCommunicator. And with the wear update will bring an update/fix to this code that will give this dialog also when not started from the GUI (also for SMSCommunicator).

It would be nice if we could keep this back until the fix is in.

AdrianLxM:
I think we will have to decide if we want to actually turn on the screen if a bolus is given via SMS/Watch.
In this case we would have to add one flag to the wakelock:

`mWakeLock = powerManager.newWakeLock(PowerManager.SCREEN_DIM_WAKE_LOCK|PowerManager.ACQUIRE_CAUSES_WAKEUP, "ConfigBuilderPlugin");`

We could also make this optional for the user.

PR #195, opened 2017-03-02
Simply truncating the carbs if user enters decimal value.  Anything l…

samihusseingit:
Simply truncating the carbs if user enters decimal value.  Anything less than 1g will be 0.

First time making an update :)


AdrianLxM:
Hi Sami,

great that you started to contribute. If we want that behaviour (I never log anything finer than 1g - not sure if others do?), there is also the posibility for the text field to not accept decimal places in the first place.
As far as I remember it is done by `android:inputType:"number"`.

cheers.

samihusseingit:
hehe DOH!.  Thank you for this.  Never did android development, gotta learn.

Thanks

PR #233, opened 2017-05-09
move evaluation of always_use_shortavg setting to the OAPS plugins

AdrianLxM:
For the overview and the watch we still may use the standard delta, even if this setting is enabled. (Also good to check that it is the same as in xDrip e.g.)


MilosKozak:
use SP.getBoolean(....)

PR #241, opened 2017-05-12
Get correct calibration value if entered with (onboard) keyboard

AdrianLxM:
Only tested in mg/dl so far.


AdrianLxM:
Should also work in mmol/l :)

PR #249, opened 2017-05-22
cancel temp on suspend loop

AdrianLxM:
@MilosKozak it would be great if you could check if my change to the SMS-Communicator makes any sense. I cant't test it.

thanks


AdrianLxM:
I've added the same to "disable loop".

PR #269, opened 2017-06-29
visualize absolute rate emulation as percentage

AdrianLxM:
A setting that depends on "Use extended boluses for >200%" and visualizes temp basal emulations with extended boluses as %age:

If enabled:
 * Shows cancel-button in overview as %age
 * Shows rate on watchface as %age
 * Keeps the long status in overview as it is so it is still visible what the pump is actually doing

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/db/TemporaryBasal.java
+++ b/app/src/main/java/info/nightscout/androidaps/db/TemporaryBasal.java
@@ -273,7 +273,7 @@ public class TemporaryBasal implements Interval {
 
     public String toStringShort() {
         if (isAbsolute) {
-            if(SP.getBoolean(R.string.key_danar_visualizeextendedaspercentage, false)){
+            if(SP.getBoolean(R.string.key_danar_visualizeextendedaspercentage, false) && SP.getBoolean(R.string.key_danar_useextended, false)){
                 Profile profile = MainApp.getConfigBuilder().getProfile();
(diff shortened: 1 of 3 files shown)


app/src/main/java/info/nightscout/androidaps/db/TemporaryBasal.java
```diff
@@ -271,6 +273,15 @@ public String toStringFull() {
 
     public String toStringShort() {
         if (isAbsolute) {
+            if(SP.getBoolean(R.string.key_danar_visualizeextendedaspercentage, false)){
```
AdrianLxM: I had tested that:
When you disable 200%+, extended boluses will not be shown as button or on the watch.
Only as treatments in the chart (cyan text).

Should I add this anyway?

MilosKozak: if you:
turn on 200+
turn on %
turn off 200+

i guess % will be still kept this way ... or am i wrong?

AdrianLxM: I think so, yes.

Just the method will not be called in that case as it is guarded by `treatmentsInterface.isTempBasalInProgress()` or `activeTemp` is another one (tempBasal + extended Bolus).

But you are right, better add this that the function can be universally called (make future changes easier).

PR #275, opened 2017-07-03
Fake extended fix

AdrianLxM:
Hi @MilosKozak,

could you please check, if this is how you thought it should be.
As you said, most was just representation and IOB not affected.

Only `TreatmentsPlugin#TreatgetTempBasalAbsoluteRateHistory()` was feeding the rate to determine-basal.

A fakeExtended rate is still an absolute rate here. I think this should remain for uploading purposes?


AdrianLxM:
Issue: Low-Temps now show as 100% :(

AdrianLxM:
... initialisation issue fixed.

PR #279, opened 2017-07-12
Cpp wear

AdrianLxM:
Allows to set  percentage and timeshift from AndroidWear.


AdrianLxM:
Thanks @Heiner1 for testing this.

Issue #282, opened 2017-07-13
Bolus wizard suggests bolus unsupported by pump


jotomo:
e.g. 0.45 U when the pump only supports 0.1 U increments, thus either 0.4 U or 0.5 U.
Should the wizard honour the pumps capabilities or should this be rounded by the pump code?

AdrianLxM:
I'd prefer having the feedback in the GUI. The pump-plugin does report the "Bolus Step".

AdrianLxM:
fixed in #285. Thanks @jotomo

PR #285, opened 2017-07-14
Bolus wizard: use PumpDescription to round final bolus to a value com…

jotomo:
…patible with the pump.

Suggested fix for #282


AdrianLxM:
Can `MainApp.getConfigBuilder().getPumpDescription()` return null, do we have to check for that?

It should not happen in practice as none of the buttons in the overview to start the bolus wizard would be active and the watchlogic (performed on the mobile part) should not allow to do this operation when the pump is not initialized.

AdrianLxM:
It will only be instantiated in those 3 cases where the pump is instantiated -> merging as @MilosKozak approved it :) 

Thanks @jotomo

PR #286, opened 2017-07-15
Temp target logic

AdrianLxM:


MilosKozak:
merge it if it works.... i cannot test it now

PR #290, opened 2017-07-17
Updating Bulgarian translation

PoweRGbg:
I've translated latest changes


AdrianLxM:
It looks like there are some translations of strings in, that are not (yet) in the default strings.xml

PoweRGbg:
Removed! :)

PR #292, opened 2017-07-17
Wear tdd

AdrianLxM:
* simplifies the menus on wear.
* combines/adds the targets and last OAPS result and OAPS reason to the loop status on wear.
* adds the TDD stats to wear.
* defaults to single target for temp targets (like on the phone) now.


MilosKozak:
@AdrianLxM feel free to merge when you decide

AdrianLxM:
thanks

PR #295, opened 2017-07-18
On resume update gui

jotomo:
@AdrianLxM First commit as discussed, second commit (extraction) as a quick experiment, untested, to see if this makes sense. WDYT?


AdrianLxM:
Looks good but not yet tested. I've added two more fragments.

PR #296, opened 2017-07-19
Align OpenAS(A)MA fragments with layout of other fragments.

jotomo:
Removes a margin that makes the top elements jump up and
down when flipping through tabs.


AdrianLxM:
:+1:

PR #299, opened 2017-07-19
NSClient quickfix

AdrianLxM:
* Removes some debug output that stresses the broadcast system.
 * Setting that allows the user to not send broadcasts to other apps. Most AAPS-Users deactivate the reception of broadcasts in xDrip anyways. Not sending them will cause less load on the system.


MilosKozak:
What is the preference for?

MilosKozak:
See it now, sorry on phone only. Do you think it should be user selectable in nsclient mode?

AdrianLxM:
With your great rework, NSClient is a good standalone viewer/client for NightScout, so I think it would make sense to have this setting in NSClient mode as well.

MilosKozak:
:+1:

PR #318, opened 2017-07-23
Nsclient remove internal broadcasts

AdrianLxM:
I refactored the code that when NSClient receives Data, it calls the local method directly and AAPS no longer listenes to NSClient data broadcasts.

As I don't use NS atm, please test before merging.

@jotomo could you test this, please?

@MilosKozak, I've noticed one thing:
[handleRemovedTreatment](https://github.com/MilosKozak/AndroidAPS/blob/dev/app/src/main/java/info/nightscout/androidaps/plugins/NSClientInternal/broadcasts/BroadcastTreatment.java#L99) receives a JSONArray but does no splitting.
[handleNewTreatment](https://github.com/MilosKozak/AndroidAPS/blob/dev/app/src/main/java/info/nightscout/androidaps/plugins/NSClientInternal/broadcasts/BroadcastTreatment.java#L45) and [handleChangedTreatment](https://github.com/MilosKozak/AndroidAPS/blob/dev/app/src/main/java/info/nightscout/androidaps/plugins/NSClientInternal/broadcasts/BroadcastTreatment.java#L74) on the other hand split the array. (in `master` none of them do splitting, it seems)


MilosKozak:
handleRemovedTreatment is always small

MilosKozak:
diff is too big to read. :(
the reason why all these different intents were handled in one service was that the received intent could lead to pump action. and it uses handler from intentservice. so these actions were not mixed and executed one by one.

we should keep this logic or replace by similar

AdrianLxM:
In this case I would opt for staying with broadcasts and not following this approach here. Otherwise it would be hard to synchronize these with broadcasts from other apps like xDrip.

Maybe use local broadcasts instead?

PR #320, opened 2017-07-24
Nsclient local broadcasts

AdrianLxM:
This time with local instead of global broadcasts.
(always app-local broadcasts, global broadcasts when activated, AAPS does not listen to global NSClient broadcasts but only the local ones)

Tested with treatments: They always reload in AAPS. If the setting is enabled, xDrip receives Treatments from NS, if disabled, not.


AdrianLxM:
Also tested an "Urgent High alert".

MilosKozak:
replaced by https://github.com/MilosKozak/AndroidAPS/pull/323

PR #340, opened 2017-08-05
Move to range in profile - WIP

MilosKozak:


MilosKozak:
@jotomo @AdrianLxM test please

AdrianLxM:
Works for me. 
Initial popup works and it removes the targets from the settings as it should.

OAPS and the Overview use the correct targets
Temp-Targets also still  work.

jotomo:
Checked all profile plugins, TT, OAPS, working good.
@MilosKozak Thanks for picking this up, got lost in all the other stuff ...

PR #344, opened 2017-08-07
Wear: add git build version to status activity

LadyViktoria:


MilosKozak:
is it the same as @AdrianLxM 's PR?

PR #348, opened 2017-08-09
disconnect15min (part 2 of 2)

gempickfordwaugh:
...because not everyone has a 30min shows ;-)

Hope this PR works!


MilosKozak:
pump does not support 15 min

AdrianLxM:
@gempickfordwaugh it seems that only part of the code (just the text) has come through.

@MilosKozak With the next SGV after the 15 minutes the loop should start again though?
This should be the same situation for 30 minutes as non-recent firmware on the pumps only allows 0% for 60 minutes.

gempickfordwaugh:
I've been using it (on master not Dev branch) for two weeks and not had any
problems, though my shower does normally last 9-13mins so I am resuming it
more than letting it kick in of its own accord.

Oh dear. Well I got halfway there!

savek-cc:
15 min is supported by other pumps though (e.g. combo)

PR #357, opened 2017-08-13
change logfile name pattern

savek-cc:
change logfile name pattern to still be rotated once a day but include the hour-minute-seconds as timestamp in the filename


savek-cc:
this results in filenames like
`/storage/emulated/0/Android/data/info.nightscout.androidaps/files/AndroidAPS._2017-08-13_14-20-07_.0.zip`

MilosKozak:
👍

PR #361, opened 2017-08-14
Oref Insulincurves

AdrianLxM:
This brings the new insulin curves from Oref0/1 to AndroidAPS.
https://github.com/openaps/oref0/pull/568/files
Thanks @sulkaharo

There will be 3 new plugins for insulin calculation:
screenshot_20170814-153356

 One for the rapid-acting insulin curve:
screenshot_20170814-153246
screenshot_20170814-153257

 One for the ultra-rapid-acting insulin curve ...
screenshot_20170814-153322

... and one with a freely (between 35 and 120 minutes) parametrizable peak time: 
screenshot_20170814-153411
screenshot_20170814-153429

The minimum DIA will be 5 hours. This corresponds to a much shorter DIA with our current model as the activity curve will get very flat at the end. The main configuration parameter will be the peak time. 

Also fixes a small bug where the insulin curves would always be plotted on top of each other (not visible as always the same) piling up data.


MilosKozak:
is there a plan to keep old insulin profile in oref?

MilosKozak:
we have to warn user there is shorter DIA than minimal. probably notification to make user acknoledge it

AdrianLxM:
At the moment the insulin plugin does show a warning (in the fragment only) and uses the minimum of 5 hours. So at least having programmed a too short DIA is not an usafe state.

screenshot_20170814-182712
(Example when 4h are programmed)

@MilosKozak do you have an idea how to add a more visible warning? A Toast like when a target is out of hard limits every time the plugin is called might be too much as it may get called for many treatments in a row.

MilosKozak:
new notification on overview without autoacknolegde

MilosKozak:
it will stay there until user change DIA

sulkaharo:
The OpenAPS changes are still ongoing; the plan is to remove as much code that deals with DIA as possible (use active insulin amounts etc instead) and potentially re-brand the whole activity time to something new, given it looks like there's a lot of users out there who are used to a very short DIA (in many cases, just 2 hours) that's obviously wrong but given the user might have had the setting at the short DIA for a long time, they assume it's correct, even in cases where the user hasn't actually used the Bolus Wizard for corrections and thus the DIA setting hasn't actually affected the corrections.

PR #363, opened 2017-08-15
Overview: format target using unit specific format.

jotomo:
Tested with mg/dl


MilosKozak:
`Profile.toUnitsString(low, low, units)` should be enough. it will pick up the correct value

jotomo:
Done

PR #365, opened 2017-08-16
Restrict DIA for expectedDelta

AdrianLxM:
In order to make the new insulin profiles that need a longer DIA compatible with the classic determine-basal versions, restrict the DIA to the old default of 3 hours.

WIP: only tested for a few loop cycles.


AdrianLxM:
Tested it on MA and AMA.
It really runs much smoother when expectedDelta is calculated as intended.

PR #367, opened 2017-08-16
show notification on toast alarm

AdrianLxM:
... this way you can see the message even if you are too late for the toast.


MilosKozak:
👍

PR #370, opened 2017-08-16
Check for tempbasal not RealTempbasal

AdrianLxM:
WIP: not tested yet.
Feel free to merge when tested.


AdrianLxM:
Tested. Now also cancels high-temps emulated with extended bolus

PR #373, opened 2017-08-17
Profile switch viewer update

AdrianLxM:


AdrianLxM:
Added a check to avoid NPEs and made some variables non-static.

PR #375, opened 2017-08-18
Delay in bg update on resume

samihusseingit:


MilosKozak:
what should this solve?

samihusseingit:
In my case milos everytime i wake up my phone while androidaps is on....it takes approx 15 or so seconds to display the updatef bg becsuse it dows all the logic first then updates the bg.  Updating the bg before all the other loop logic makes it more real time

PR #377, opened 2017-08-19
Only warn about invalid DIA if coming from an active insulin plugin.

jotomo:
Since the id of the insulin plugin used is saved with each treatment record,
calculating IOB will ask such a plugin. If another insulin plugin is active now,
this would still raise an error about a too short DIA, which is confusing at best.
This change only displays this warning if the plugin is enabled.

@AdrianLxM

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/InsulinOrefCurves/InsulinOrefBasePlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/InsulinOrefCurves/InsulinOrefBasePlugin.java
@@ -18,6 +18,9 @@ public abstract class InsulinOrefBasePlugin implements PluginBase, InsulinInterf
 
     public static double MIN_DIA = 5;
 
+    protected static boolean fragmentEnabled = false;
+    protected static boolean fragmentVisible = false;
+
     long lastWarned = 0;
 
     @Override
@@ -51,7 +54,7 @@ public abstract class InsulinOrefBasePlugin implements PluginBase, InsulinInterf
         if(dia >= MIN_DIA){
             return dia;
         } else {
-            if((System.currentTimeMillis() - lastWarned) > 60*1000) {
+            if(fragmentEnabled && (System.currentTimeMillis() - lastWarned) > 60*1000) {
                 lastWarned = System.currentTimeMillis();
                 Notification notification = new Notification(Notification.SHORT_DIA, String.format(MainApp.sResources.getString(R.string.dia_too_short), dia, MIN_DIA), Notification.URGENT);
                 MainApp.bus().post(new EventNewNotification(notification));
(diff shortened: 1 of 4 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/InsulinOrefCurves/InsulinOrefBasePlugin.java
```diff
@@ -18,6 +18,9 @@
 
     public static double MIN_DIA = 5;
 
+    protected static boolean fragmentEnabled = false;
```
AdrianLxM: Shouldn't this be in the child and not in the base plugin?
Wouldn't having it here enable all derived plugins at once?

jotomo: I thought I had thought it through, but I was thinking too much ... or just coding half-asleep ;-)
It would work like this only for non-static fields.
The only required (replacing this commit) should be to add _isEnabled(INSULIN)_ to the condition determining whether to issue a warning. Does that sound right?

AdrianLxM:
I didn't realise such a behaviour when adding the curves. @jotomo could you please point me to the code where it does that.
What I also don't understand is why it would rise an alarm. The DIA is passed from outside.

@jotomo, @MilosKozak, this leads to the questions:

1. Do we really want to use an "old" insulin plugin that the user discarded/changed?

2. For profile we always seem to calculate on the current one - not the one when the treatment was created
https://github.com/MilosKozak/AndroidAPS/blob/f703c82ca7469022c824f1f9098a7b0afd891040/app/src/main/java/info/nightscout/androidaps/plugins/Treatments/TreatmentsPlugin.java#L174
Is this the correct behaviour? (If the one from the time when it was created was used, it indeed would continue warning about the short DIA as long as IOB gets considered).

jotomo:
> I didn't realise such a behaviour when adding the curves. @jotomo could you please point me to the code where it does that.

In _Treatment.java_, via _insulinInterfaceID_:
```java
    public Iob iobCalc(long time, double dia) {
        if (!isValid)
            return new Iob();
        InsulinInterface insulinInterface = MainApp.getInsulinIterfaceById(insulinInterfaceID);
        if (insulinInterface == null)
            insulinInterface = ConfigBuilderPlugin.getActiveInsulin();

        return insulinInterface.iobCalcForTreatment(this, time, dia);
    }
```

AdrianLxM:
@jotomo I found the position in the code. IMHO this is not the correct behaviour.
It also prohibits the user from doing changes and directly seeing the effect. we need to rewrite the piece of code treating IOB calc for treatments anyway in order to get the bolussnooze working again.

AdrianLxM:
In addition to that: Now I think I know where the warning comes from:

It uses the Insulin-Interface from the past (by ID) that might allow only DIAs >5h. On the other hand, the DIA passed through to the interface is always from the currently active profile.
https://github.com/MilosKozak/AndroidAPS/blob/f703c82ca7469022c824f1f9098a7b0afd891040/app/src/main/java/info/nightscout/androidaps/plugins/Treatments/TreatmentsPlugin.java#L178

This is one reason more to always use the current insulin interface in the first place and not use historic ones.

Instead of masking/hiding the warning I therefore would change the logic behind it and not let this situation happen in the first place.

PR #378, opened 2017-08-19
Fix initalizing profile with default values.

jotomo:
@MilosKozak Noticed this hunting another issue, seems like a C&P bug? Not sure of the implications, but the current versions seems wrong :-)


MilosKozak:
good catch, thanks!

Issue #376, opened 2017-08-19
Bolussnooze


AdrianLxM:
The amount of insulin considered as bolussnooze does not change, even If the insulin-plugin or the bolusznooze dia divisor is changed.

@jotomo

AdrianLxM:
It looks like a double issue:
1. Is the dia-divisor fetched correctly from the settings? It uses `SP#getDouble(String key, Double defaultValue` instead of `SP#getInt(int resourceID, Integer defaultValue)` despite `openapsama_bolussnooze_dia_divisor` being a key in the strings. Still looks ok but we should switch to a unified way of fetching things.

2. The main issue:
 * Shortening the DIA does no longer work with the new curves as they no longer only get stretched by the DIA but also have a peak time fetched from the settings.
 * We must now multiply the time passed since the Bolus happened. (in `Treatment#iobCalc()`)

PR #380, opened 2017-08-19
Dia divisor / bolussnooze fix

AdrianLxM:
This should fix the issues we have around the dia divisor and bolus snooze.

It fixes Issue https://github.com/MilosKozak/AndroidAPS/issues/376
The idea behind the "dia divisor" was that the time for the bolussnooze passes faster for a treatment than for IOB to decay the bolussnooze faster. With the bilinear curves this was done by dividing the DIA what no longer can be done with the new curves. Therefore the new implementation uses the same DIA but actually lets the time pass faster.
This is the same way of implementing it as in OpeAPS-Oref0/1:
https://github.com/openaps/oref0/pull/568/files#diff-b43ed8995d69a77e23443ef00bb36bb8R140

It should also fix the issue of "old" plugins being used with the state (dia, peak, ...) of the new ones, leading to "alarms from the past" and that the user did have no appropriate feedback on parameter change - as described in https://github.com/MilosKozak/AndroidAPS/pull/377

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/db/Treatment.java
+++ b/app/src/main/java/info/nightscout/androidaps/db/Treatment.java
@@ -183,10 +183,8 @@ public class Treatment implements DataPointWithLabelInterface {
     public Iob iobCalc(long time, double dia) {
         if (!isValid)
             return new Iob();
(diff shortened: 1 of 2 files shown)


jotomo:
LGTM, reported bolus snooze looks much better now and reacts to changes in the divisor.

PR #389, opened 2017-08-22
Cleanups

jotomo:


app/src/main/java/info/nightscout/androidaps/db/Treatment.java
```diff
@@ -50,32 +50,9 @@
     @DatabaseField
     public boolean isSMB = false;
 
-    @DatabaseField
```
MilosKozak: I'd prefer to keep it. It's not done yet but it's prepared for supporting multiple insulin types with MDI or pump+MDI

MilosKozak: Anyway by doing this it would be needed to increase database sequence number

jotomo: I was assuming it was obsolete now after Adrian changed the insulin curves logic. So only keep the database fields, or also leave constructors and their usage unchanged?

MilosKozak: Constructor is not an issue but database schema is. I prepared maximum possible inside this huge release to prevent recreating table in the future.
The idea is to fill pluginID and dia at time of creation and use these values from `Treatment` class.
@AdrianLxM is using different insulin types. maybe he can finish it and allow to enter another treatments with different insulin types for example from careportal

AdrianLxM: I would just keep the field to preserve DB integrity for now.

When this comes we will need to add other fields as well the internal configuration of the insulin plugin will depend on more. Especially if we add another plugin for diaport users.

jotomo:
Reverted schema changes, didn't think of that.

_Sent from my ZTE A2017G using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #411, opened 2017-09-02
WIP: Timelistedit performance

jotomo:
Not sure exactly what's going on, but with the _timeSpinner.setOnItemSelectedListener_ callbacks in place, the GPU is maxed out redrawing the fragment, preventing me from editing anything. This only occurs on a newer, faster phone of mine.

### DIFF
--- a/app/src/main/java/info/nightscout/utils/TimeListEdit.java
+++ b/app/src/main/java/info/nightscout/utils/TimeListEdit.java
@@ -1,16 +1,13 @@
 package info.nightscout.utils;
 
 import android.content.Context;
-import android.os.Build;
 import android.support.v4.content.ContextCompat;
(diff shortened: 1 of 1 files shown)


MilosKozak:
how did you replace the code inside `setOnItemSelectedListener`? It stores data

jotomo:
@MilosKozak I call it _experimental learning_: remove something and see what breaks ;-)
Seriously though, new callbacks are created on each run of _buildView_ and it appears they all stay active and get called. Best guess so far: since the controls receive a new ID on every call of the method, the previous callback is not overridden if the ID differs. If the _View.generateViewId_ method happens to regenerate the same ID on my older device, but a different one on my newer phone, then that would explain it. I still need to learn a lot about Android :-)
Either way, I'll close this for now since it's not ready.

PR #415, opened 2017-09-09
TT in wizard

jotomo:


app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/WizardDialog.java
```diff
@@ -381,31 +405,17 @@ private void initDialog() {
...
-            Double targetBGHigh = profile.getTargetHigh();
-            Double bgDiff;
-            if (lastBgValue <= targetBGLow) {
-                bgDiff = lastBgValue - targetBGLow;
-            } else {
-                bgDiff = lastBgValue - targetBGHigh;
-            }
-
-            bg.setText(lastBg.valueToUnitsToString(units) + " ISF: " + DecimalFormatter.to1Decimal(sens));
-            bgInsulin.setText(DecimalFormatter.to2Decimal(bgDiff / sens) + "U");
```
jotomo: _bgInsulin_ is calculated and set in _calculateInsulin()_,  from the result of calling _BolusWizard_ and _calculateInsulin_ is called an the end of this method , so this can be removed.

app/src/main/java/info/nightscout/utils/BolusWizard.java
```diff
@@ -108,13 +111,13 @@ public Double doCalc(Profile specificProfile, Integer carbs, Double cob, Double
         }
 
         // Total
-            calculatedTotalInsulin = totalBeforePercentageAdjustment = insulinFromBG + insulinFromTrend + insulinFromCarbs + insulingFromBolusIOB + insulingFromBasalsIOB + insulinFromCorrection + insulinFromSuperBolus + insulinFromCOB;
-
-            //percentage
-            if(totalBeforePercentageAdjustment > 0){
-                calculatedTotalInsulin = totalBeforePercentageAdjustment*percentageCorrection/100d;
-            }
+        calculatedTotalInsulin = insulinFromBG + insulinFromTrend + insulinFromCarbs + insulingFromBolusIOB + insulingFromBasalsIOB + insulinFromCorrection + insulinFromSuperBolus + insulinFromCOB;
```
jotomo: @AdrianLxM This is unrelated to the change of this commit, but I really dislike assignments to multiple variables on a single line. The variable assigned to _calculatedTotalInsulin_, is a nice abstraction of whatever calculation is performed on the right, so when reading the code I often just read the name of the variable, not the details that make it up (until needed). That makes it easy to miss that the 'calculation' contains another assignment. This change results in more code, but is IMO easier to read. WDYT?

app/src/main/java/info/nightscout/utils/BolusWizard.java
```diff
@@ -50,23 +47,29 @@
...
         this.superBolus = superBolus;
         this.trend = trend;
 
-
         // Insulin from BG
         sens = specificProfile.getIsf();
         targetBGLow = specificProfile.getTargetLow();
         targetBGHigh = specificProfile.getTargetHigh();
+        if (tempTarget != null) {
+            targetBGLow = tempTarget.low;
```
MilosKozak: must be converted to mmol if needed

jotomo: Fixed and tested

jotomo:
Hm, should the TT option be visible depending on a preference, like the superbolus option? With SMB seting an _Eating Soon_ target will already bolus for it.

PoweRGbg:
Mealbolus will also enable SMB, so it will be enabled anyway :)

MilosKozak:
did someone test it? @AdrianLxM @PoweRGbg

jotomo:
@MilosKozak I've been usling it for a week+ on my production system (mg/dl), but I'm not aware of anyone else testing this.

PR #419, opened 2017-09-17
Time Spinner Local Profile Fix

dylanleonard:
When a time later than 12:00 PM was chosen in the Local Profile fragment (and anywhere else toSeconds is called), the AM version was being returned instead by toSeconds in DateUtil

### DIFF
--- a/app/src/main/java/info/nightscout/utils/DateUtil.java
+++ b/app/src/main/java/info/nightscout/utils/DateUtil.java
@@ -85,15 +85,15 @@ public class DateUtil {
     }
 
     public static int toSeconds(String hh_colon_mm) {
-        Pattern p = Pattern.compile("(\\d+):(\\d+)( a.m.| p.m.|)");
+        Pattern p = Pattern.compile("(\\d+):(\\d+)( a.m.| p.m.| AM | PM)");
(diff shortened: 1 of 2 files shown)


app/src/main/java/info/nightscout/utils/DateUtil.java
```diff
@@ -85,15 +85,15 @@ public static Date toDate(Integer seconds) {
...
         Matcher m = p.matcher(hh_colon_mm);
         int retval = 0;
 
         if (m.find()) {
             retval = SafeParse.stringToInt(m.group(1)) * 60 * 60 + SafeParse.stringToInt(m.group(2)) * 60;
-            if (m.group(3).equals(" .a.m") && m.group(1).equals("12"))
+            if (m.group(3).equals(" AM") && m.group(1).equals("12"))
                 retval -= 12 * 60 * 60;
-            if (m.group(3).equals(" p.m.") && !m.group(1).equals("12"))
+            if (m.group(3).equals(" PM") && !(m.group(1).equals("12")))
```
MilosKozak: on my android ".a.m" and ".p.m" is used 
add `||` here to support both variants

PR #426, opened 2017-09-22
LoopPlugin: adjust negative requested rates to 0% rate.

jotomo:
This adjusts the oref0 output to make it useful input for AAPS, rather than handling a negative rate as constraint violation and fixing the issue in adjusting a negative rate.


jotomo:
I'm still testing this to make sure there are no side effects, but thought I'd put it here for discussion.

PR #434, opened 2017-09-29
alesinka's icons added to the DanaR tab

PoweRGbg:
Added @alesinka icons to the DanaR tab and fixed the problem with the filenames


PoweRGbg:
wrong again

PR #436, opened 2017-09-30
Watchface changes

Heiner1:
Some watchface changes for better display tested on SW3

### DIFF
--- a/wear/src/main/java/info/nightscout/androidaps/watchfaces/BaseWatchFace.java
+++ b/wear/src/main/java/info/nightscout/androidaps/watchfaces/BaseWatchFace.java
@@ -32,6 +32,7 @@ import com.ustwo.clockwise.common.WatchShape;
 import com.ustwo.clockwise.wearable.WatchFace;
 
 import java.util.ArrayList;
+import java.util.Date;
 
 import info.nightscout.androidaps.R;
 import info.nightscout.androidaps.data.BasalWatchData;


AdrianLxM:
I really prefer white on dark as it is much better readable in sunlight.
How about a setting to toggle that?

Heiner1:
That's again a question of more comfort or less work; if I would have more time just now, I'd try to find out how to integrate a switch button or s.th. else; but it's not very necessary, and we could let it on the to do list, what do you think?

Heiner1:
watchface_nochart.png (PNG-Grafik, 280 × 280 Pixel) (all others have 280x280 pixel)  is the most important to merge, because the original watchface 320x320Pixel doesn't match with e.g. the SW3

AdrianLxM:
@Heiner1 you can change the text size in your android wear settings. The text-size you have set now will make the delta too small to read on my watch.

The other thing I'd rather not have is real looking numbers as placeholders. Can you confirm, that `tools:text` will only set the number in the layout preview in Android-Studio but not on the watch itself when no data is present?
What is shown on the watchface without any placeholder? Are the settings still possible to open on double-tap?

Heiner1:
@AdrianLxM  OK I'll check text size, but I know it's small, wanted to have more space for BG Graph; and I think about, if it's more efficient to see an delta arrow instead of numbers, because a watch is to recognize fast what's the most important - like in xDrip

AdrianLxM:
I on the other hand believe a trend arrow to be pretty much useless if I have the delta and average delta.
I'd have to map the arrow to a number anyways before evaluating it - and then -4mg/dl and +4mg/dl, what is a huge difference, would still look the same.
That's why I created the big-chart watchface in the first place: I didn't need the arrow so I had more space for the chart by not showing it.

I know others get along better with symbols like arrows than with numbers.

What do you think of creating a new watchface? It does not take up much ressources on the watch.
You can copy the Java-File; register the copied one in the manifest; copy the layout and in the onCreate()-method of the new watchface, just load the copied layout.
From there you can make any modifications.

Heiner1:
I think you're right, it makes sense to create another watchface. I think it's not good to have too much choices, but maybe some are interested and profit by this. I hope to manage it to create a new one or two. So we should close the PR?

Issue #437, opened 2017-10-03
Pump disconnect -> no 0% Temp


McHoffi:
When disconencting Pump, AAPS won't set a 0% Temp, resulting in wrong IOB.

stavlor:
If your flagging it as a disconnect it will automatically set a 0.0u/hr temp

MilosKozak:
@McHoffi it works well for me ... do you have more info?

McHoffi:
![screenshot_20171004-110428](https://user-images.githubusercontent.com/24695277/31168847-144f00d6-a8f7-11e7-9238-10a4e5515715.png)
I don't know what you need, a logfile? On the screenshot you see, that it is disconnected, but not 0 temping.

AdrianLxM:
@McHoffi just to make sure: you presssed "Disconnect..." not "Suspend..."

Suspending the loop means there will be no automatic looping for a certain amount of time. You are in control of your pump.

Disconnecting means, you detach the pump from your body. No insulin will go into your body no matter what -> 0% temp.

AdrianLxM:
Probably a log file would not be a bad idea :)

It might be a situation where a loop-calculation is already started (setting it to a neutral temp as result?) just in the moment when you press that button?

Heiner1:
Yesterday I saw the same but now I tried it again with all timing possibilities and it worked; maybe @AdrianLxM  you're right and it's a bad coincidence with a background action;

PR #438, opened 2017-10-04
Fix zerotemp misinterpretation

AdrianLxM:
This fixes https://github.com/MilosKozak/AndroidAPS/issues/437 for me.

As with this commit https://github.com/MilosKozak/AndroidAPS/commit/dc6cbadbdf8eb3799e37f5f1e14beea55eb77f7c a lot of history-queries have been introduced it might be good to look at them as well.


AdrianLxM:
Should be targeting dev not master.

PR #445, opened 2017-10-06
update NP value after manual entry

AdrianLxM:
Fixes the problem that after entering a number with the keyboard, the value was not updated and therefore +/- buttons used an old internal value if pressed afterwards.


PoweRGbg:
Great fix - it was really annoying

MilosKozak:
does it really works on `this` and not on `editText` ?

AdrianLxM:
@MilosKozak 

It passes it through to editText:

```
     public void addTextChangedListener(TextWatcher textWatcher) {
         editText.addTextChangedListener(textWatcher);
}
```

PR #446, opened 2017-10-06
overview checkboxes cutoff fix

AdrianLxM:
Do not cut off the head of the checkboxes in the overview :) 

Once you see it.... ;)


MilosKozak:
there is one more fragment

AdrianLxM:
I raise to two more and: done :)

Thanks, I didn't think of that NS-Client could also have those.

PR #451, opened 2017-10-09
Overview: align checkboxes and labels vertically.

jotomo:
Before:
image

After:
image


jotomo:
Changes are copied from main fragment to all others, tested in emulator with all four variants.

PR #453, opened 2017-10-10
profile on wear

AdrianLxM:


app/src/main/java/info/nightscout/androidaps/plugins/Careportal/Dialogs/NewNSTreatmentDialog.java
```diff
@@ -722,11 +722,11 @@ public static void doProfileSwitch(final int duration, final int percentage, fin
             @Override
             public void run() {
                 ProfileSwitch profileSwitch = MainApp.getConfigBuilder().getProfileSwitchFromHistory(System.currentTimeMillis());
-                if (profileSwitch == null) {
+                if (profileSwitch != null) {
```
MilosKozak: the idea here was use ProfileSwitch and change % and shift only.
with this i'd remove `ProfileSwitch profileSwitch = MainApp.getConfigBuilder().getProfileSwitchFromHistory(System.currentTimeMillis());` and `if`

AdrianLxM: Sorry, this time it was me pushing late.

I put everything into the `if´ clause as it should only be able to do a background switch from only percentage and timeshift, if a current profileSwitch exists.

Do you think it is ok like it is now @MilosKozak?

MilosKozak: do we need to check for existing profileswtich?

AdrianLxM:
Closing in favour of: https://github.com/MilosKozak/AndroidAPS/pull/454
The loading speedups do have a bug and should be a different PR.

PR #462, opened 2017-10-15
Profile rhino

AdrianLxM:
For testing.

MA and AMA should now work with Rhino. I still left the old lib in for easier merges with other branches. Will remove when tested.


MilosKozak:
no need to wait. it works well

PR #466, opened 2017-10-16
Only display Extended Bolus tab in Treatments if pump supports them.

jotomo:


AdrianLxM:
What about extended boluses done from another pump (when you switch between two) or entered manually via the care-portal?

MilosKozak:
i'm not sure what's better

jotomo:
The tab could be shown depending on presence of extended bolus history data, but then it would need to be shown again when data comes in from external sources. This patch was meant for an upcoming pump that might not have extended bolus support at the beginning, but that's probably not a real use case, so we needn't further bother with what is basically a cosmetic change not affecting more than one user at this point.

PR #467, opened 2017-10-16
ConfigBuilder: hide disabled plugins by default.

jotomo:


AdrianLxM:
Does this mean when you enable a plugin, you have to also make it visible?
OAPS should be visible by default if enabled unless the user does actively disable the view.
Users should look at those tabs!

jotomo:
Plugins are shown when they're enabled. The `SMS communicator` plugin was disabled by default but set visible, so a blank tab was shown. Fixing that, I updated others as well. This fixes the initial state on first application start.

MilosKozak:
when user enable new plugin it's set as visible too by default. when plugin is disabled it should not be possible to make it visible. or am I wrong?

jotomo:
When a plugin is disabled it is set hidden and and the _show_ button is disabled. This PR only changes the initial state to prevent an inactive plugin to be shown on first start. When enabling/disabling such a bad state is not entered.

PR #485, opened 2017-10-25
Start KeepAliveService after pump init.

jotomo:


AdrianLxM:
At the moment the method `startKeepAliveService()` also starts the execution services for the pumps
Shouldn't they be started before doing the first read cyle from the pump?
I guess starting them would need to be moved up.

What is the reason behind not starting the keep alive before everything else?

jotomo:
_KeepAliveReceiver_ may or may not trigger an instant call to _Pump.refreshDataFromPump_, which then runs before the explicit initialization ..

MilosKozak:
if touch the services i'd hook starting pump services to `setEnabled`

AdrianLxM:
Included in: https://github.com/MilosKozak/AndroidAPS/pull/486

jotomo:
Resolved via  #486.

PR #489, opened 2017-10-27
NSTreatments NPE fix

AdrianLxM:
When the dialog was open and the phone goes to sleep/lock screen, Android may garbage collect and options is no longer available. -> NPE

I think as quick fix it is ok to not show a dialog at all. If the user stopped using the dialog and lets the phone sleep for a longer while with the phone going to sleep... he can open it again, if he wants it.


MilosKozak:
so simple? returning null prevents recreating dialog?

AdrianLxM:
It might create an empty dialog that is "see through". So clicking anywhere should close it. At least on my system.

I tried dismissing the dialog but that is not allowed there.

There might be a more elegant way, but this already prevents the crash.

PR #490, opened 2017-10-28
Show remaining time of TT

AdrianLxM:
screenshot_20171028-021330


MilosKozak:
nice. what about to move it to `DateUtil` like 'DateUtil::minago`
and use it in `DanaRFrangment` for last bolus time and maybe somewhere else too?

AdrianLxM:
That's a good idea.

AdrianLxM:
Done

PR #504, opened 2017-11-24
Treatments tab: display extended bolus tab if active pump supports it…

jotomo:
… or if there are any extended bolus records.


jotomo:
@AdrianLxM How about that than? Shoud cover all use cases, no?

AdrianLxM:
Looking at the code: Looks good to me :)

PR #505, opened 2017-11-25
Remove sync to pump pref, always sync

jotomo:


jotomo:
Tested with and works with the Combo.

MilosKozak:
👍

jotomo:
Added another commit - missed a spot in `NSProfilePlugin`.

MilosKozak:
what with pumps not supporting profile update?

jotomo:
That would only concern pumps that are added later on, since all currently supported pumps support profile update, right? They'd have to show a request to the user to manually update the profile, like you suggested.

AdrianLxM:
@MilosKozak  with the Combo v2 driver, all real pumps do support setting profiles (with the Combo at least if the change is not too huge and it doesn't need too much time).

The two "virtual" pumps (virtual and MDI) have a method like this:
```
    @Override
    public int setNewBasalProfile(Profile profile) {
        // Do nothing here. we are using MainApp.getConfigBuilder().getActiveProfile().getProfile();
        return SUCCESS;
}
```

PR #515, opened 2017-12-02
Only do pump alarms in APS mode

AdrianLxM:


andrew-warrington:
Tested with NSClient; is working.

PR #519, opened 2017-12-03
Update of watchfaces

andrew-warrington:
OpenAPS time since last loop now calculated on the watch rather than sent with status.


andrew-warrington:
@AdrianLxM, with this PR I have changed the openApsStatus to be a timestamp, so that the number of minutes since last loop increments automatically onTimeChanged() on the watch. This is the last change I need to make for the cockpit / AAPSv2. The app has been running without crashes for 24 hours now, so I think whatever was causing the NPE on the null intent is gone now. Do you think it is ready to merge with dev? If not, let me know what I should do next to make it ready. All the best, Andrew

PR #525, opened 2017-12-05
Added profile's target BG in NSClient mode

PoweRGbg:


app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
```diff
@@ -976,7 +976,13 @@ else if (lastBG.valueToUnits(units) > highLine)
             tempTargetView.setVisibility(View.VISIBLE);
         }
         if ((Config.NSCLIENT || Config.G5UPLOADER) && tempTarget == null) {
```
MilosKozak: looking at code above this. would not be enough to remove this condition?

PoweRGbg: Yes it'll be enough just to remove the if statement

PR #535, opened 2017-12-10
Queue: don't show progress dialog for SMBs.

jotomo:


RoumenGeorgiev:
And only if there was something to set isSMB to true :+1:

MilosKozak:
this is different in SMB branch. i'd better not merge it here

jotomo:
I'll rewrite and create a PR against SMB.

PR #538, opened 2017-12-12
changed AAPS(Large) watchface to make it better readable for "elderly…

Heiner1:
…" including green for midcolor, which affects also the other watchfaces;


Heiner1:
Decided to prepare first a color switch or separate watchface

PR #539, opened 2017-12-12
Apply bolus constraint before issuing the bolus

jotomo:


jotomo:
@MilosKozak Good point about conflicts. I've integrated this into the combo branch already though, so I'll have to deal with conflicts anyway. Would it be okay if I merge `dev` into `smb060` after this and take care of the conflicts?

MilosKozak:
yes please!

PR #548, opened 2017-12-19
Use autosensdata for wizard instead of  APSResult

AdrianLxM:
There were multiple issues on FB reportet regarding the Wizard not showing COB at first.
25593671_1398587443603919_1266871076_n
25530305_1398587470270583_1591876266_n

I don't use COB at all, so I can't test it. Please see this only as an idea what could be wrong.

@savek-cc, does that help?

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/WizardDialog.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/WizardDialog.java
@@ -53,6 +53,8 @@ import info.nightscout.androidaps.db.TempTarget;
 import info.nightscout.androidaps.events.EventNewBG;
 import info.nightscout.androidaps.events.EventRefreshOverview;
 import info.nightscout.androidaps.plugins.ConfigBuilder.ConfigBuilderPlugin;
+import info.nightscout.androidaps.plugins.IobCobCalculator.AutosensData;
+import info.nightscout.androidaps.plugins.IobCobCalculator.IobCobCalculatorPlugin;
 import info.nightscout.androidaps.plugins.Loop.LoopPlugin;
 import info.nightscout.androidaps.plugins.OpenAPSAMA.OpenAPSAMAPlugin;
 import info.nightscout.androidaps.plugins.OpenAPSMA.events.EventOpenAPSUpdateGui;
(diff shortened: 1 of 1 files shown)


savek-cc:
First test seems fine, will continue to observe.

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/WizardDialog.java
```diff
@@ -439,7 +441,7 @@ private void initDialog() {
         basalIobInsulin.setText(DecimalFormatter.to2Decimal(-basalIob.basaliob) + "U");
 
         // COB only if AMA is selected
-        if (ConfigBuilderPlugin.getActiveAPS() instanceof OpenAPSAMAPlugin && ConfigBuilderPlugin.getActiveAPS().getLastAPSResult() != null && ConfigBuilderPlugin.getActiveAPS().getLastAPSRun().after(new Date(System.currentTimeMillis() - 11 * 60 * 1000L))) {
+        if (ConfigBuilderPlugin.getActiveAPS() instanceof OpenAPSAMAPlugin) {
```
jotomo: How about getting AutosenseData here already and checking asData.cob > 0 as to wether display this or not. That would remove dependency on a specific OpenAPS plugin.

AdrianLxM: This was the next step, I was actually thinking about that. I thought this would better fit the 0.6.0 scope where COB calculation may change.

jotomo: This changes behaviour though: previously the COB line was only displayed when COB > 0. Now it's displayed all the time. Though I found the lack of the COB line irritating at times. How about disabling the COB checkbox when COB == 0, similar to check TT checkbox which is only enabled if a TT is active? That would explicitly show that no COB is active.

jotomo:
COB in wizard nicely matches COB on overview with this.

MilosKozak:
it's better to do it this way anyway. probably at time of writing BW IobCob plugin didn't exist yet

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/WizardDialog.java
jotomo: This code also duplicated in line 150.

PR #566, opened 2017-12-27
RS create bond

AdrianLxM:


MilosKozak:
does this help to connection?

AdrianLxM:
This seems to successfully "pair" the RS with my phone (visible as paired device in system bt settings) and stops the crash from happening.

PR #581, opened 2018-01-04
request new data on preference change

AdrianLxM:
@andrew-warrington, could you please check, if this works?


andrew-warrington:
@AdrianLxM, that works very well! The logic for managing visual cues for old data is unaffected as well. I tested on three of the BaseWatchFace class watch faces.

PR #587, opened 2018-01-06
Steampunk watch face: Chart readability improvement

andrew-warrington:
When the chart is highly zoomed in (1 or 2 hour timeframe), the glucose points are further apart and the small point size makes them hard to see. It seems that:

- for a zoomed out chart, smaller points and lines are easier to read, as there is more data crowded into the space

- for a zoomed in chart, larger points and lines are easier to read, as there are fewer of them on the screen and they need to stand out more.

In this commit I have indicated that point (and line) size should double normal when the chart timeframe is zoomed in to 1 or 2 hours. 

I am open to comments or suggestions.


AdrianLxM:
As the line-size of the temp basals depends on the pointsize as well, it must probably be adjusted in the BgGraphBuilder (line 149)

`                lines.add(tempValuesLine(twd, (float) minChart, factor, false, highlight?(pointSize+1):pointSize));
`

andrew-warrington:
In experimenting with it, I found that I like having the line size increase in tandem with the point size. I also find the temp basal line sizes slightly too thin when highly zoomed in.

AdrianLxM:
Does the base-basal line also depend on the point-size? Can you confirm that it also increases?

andrew-warrington:
Honestly from testing I am not 100% sure. What I can confirm is, it just looks right at all zoom levels with this modification.

Will be a few days before I can get back in front of the code to answer factually about the dependency.

andrew-warrington:
Looking at the code (line 149 you mentioned) I can confirm it increases. The change is very subtle when looking at it in testing on the watch face, but I confirm that at least for me it looks good at all zoom levels.

AdrianLxM:
Thank you @andrew-warrington

PR #597, opened 2018-01-10
Add a unit test for Insulin Plugin

triplem:
This PR adds a unit test (an easy one) for the InsulinOrefBasePlugin. 

This unit test strongly suggest the usage of dagger2 (Dependency Injection Framework), since the unit test inherits from the InsulinOrefBasePlugin, which should not be really the case. Properties in this Class are retrieved from other classes (eg. MainApp). Some of those dependencies are already factored out into their own methods, to be able to reset those to testable values.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/data/Iob.java
+++ b/app/src/main/java/info/nightscout/androidaps/data/Iob.java
@@ -12,4 +12,26 @@ public class Iob {
         activityContrib += iob.activityContrib;
         return this;
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        Iob iob = (Iob) o;
+
+        if (Double.compare(iob.iobContrib, iobContrib) != 0) return false;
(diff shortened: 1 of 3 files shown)


triplem:
The testIobCalcForTreatment is not finished yet. There should be several more tests for a result.

triplem:
There anywhere a description of the algorithm and the expected results which I could use for the tests?

jotomo:
@triplem IIRC @AdrianLxM based the code on this (and maybe stuff referenced there): https://github.com/openaps/oref0/pull/568

triplem:
@jotomo sure, me is more up to a sort of list, which shows expected results for given input parameters. Mainly for edge cases.

jotomo:
Maybe this https://github.com/openaps/oref0/blob/master/tests/iob.test.js or ask Scott?

MilosKozak:
@triplem can I merge this?

triplem:
This is just a first try and should be extended, but can be merged, yes.

triplem:
Enhanced the test slightly. Now it is finished and ready to merge.

PR #600, opened 2018-01-11
Ignore Extra translations (on dev branch)

AdrianLxM:


MilosKozak:
just for info , what is does?

AdrianLxM:
It will allow the apk to be built in "release", even if in a language like Dutch are strings that are not in English. Like in "debug".

MilosKozak:
would not be better to solve it?

PR #616, opened 2018-01-24
Overview: make checkbox labels toggle state as well.

jotomo:
@AdrianLxM Something for your jelly


AdrianLxM:
Great!

We do have different layouts for the overview fragment though. there is one for NSClient and one for smaller layouts that is missing the ID now.

jotomo:
Of course, totally forgot about those.

jotomo:
Updated and tested with all overview variants.

PR #626, opened 2018-01-29
Refactor FoodPlugin 

triplem:
Refactor the FoodPlugin into a totally encapsulated structure.


triplem:
Found one bug, need to fix this one tomorrow.

app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
```diff
@@ -219,7 +236,7 @@ public void resetDatabases() {
             TableUtils.createTableIfNotExists(connectionSource, ExtendedBolus.class);
             TableUtils.createTableIfNotExists(connectionSource, CareportalEvent.class);
             TableUtils.createTableIfNotExists(connectionSource, ProfileSwitch.class);
-            foodHelper.resetFood();
+//            resetFood();
```
MilosKozak: This has to be handled some way

app/src/main/java/info/nightscout/androidaps/plugins/Food/FoodFragment.java
```diff
@@ -121,7 +121,7 @@ public void afterTextChanged(Editable s) {
                 }
             });
 
-            RecyclerViewAdapter adapter = new RecyclerViewAdapter(MainApp.getDbHelper().foodHelper.getFoodData());
+            RecyclerViewAdapter adapter = new RecyclerViewAdapter(MainApp.getSpecificPlugin(FoodPlugin.class).getService().getFoodData());
```
MilosKozak: `MainApp.getSpecificPlugin` will be preffered way or `FoodPlugin.getPlugin()` ?

app/src/main/java/info/nightscout/androidaps/plugins/Food/FoodService.java
```diff
@@ -0,0 +1,361 @@
...
+            }
+        } catch (JSONException e) {
+            log.error("Unhandled exception", e);
+        }
+    }
+
+    public void createFoodFromJsonIfNotExists(Food food) {
+        try {
+            this.getDao().createOrUpdate(food);
+        } catch (SQLException e) {
```
MilosKozak: `scheduleFoodChange();` should be called on every change

triplem: I will rework this one, so that the createOrUpdate is called inside the service, which will call scheduleFoodChange(). This was the bug I was takling about. 
Thanks for checking the code ;-)

app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
triplem: It is handled now ;-) Thanks for the hint.

app/src/main/java/info/nightscout/androidaps/plugins/Food/FoodFragment.java
triplem: Right now, I am using MainApp.getSpecificPlugin, in a later phase, I would like to handle this via the PluginRegistry.

triplem:
The DAO is now called only readonly without the call to scheduleFoodChange(). I created C(R)UD (Create Read Update Delete) methods which delegate to the DAO only. In the Create/Update/Delete methods the corresponding scheduleFoodChange() is called.

MilosKozak:
fixed small bugs and merged ... what next? :)

PR #627, opened 2018-01-29
Watchfaces

andrew-warrington:
BaseWatchFace: Removal of ListenerService calls in order to not conflict with Bolus and Wizard activities called by MainMenuActivity.


MilosKozak:
ready to merge?

PR #632, opened 2018-01-31
Add Insight Pump Plugin

jamorham:
**This PR adds support for the Insight pump.**

TBRs, Boluses, Extended Boluses, History and Status reading have been tested with two devices and delivery and reporting was as expected during those tests. 

A slightly modified AAPS was used with `current_basal_safety_multiplier` set to max `2.5` (only integers are currently possible via the preferences) this was tested in open and closed loop mode over a period of days and TBRs appeared sensible and were verified on the pump screen.

_The Insight will raise a pump alarm if a TBR is cancelled, so when a TBR cancellation is requested it instead sets the rate to 90% for 15 minutes. I estimate this means at worst a 2.5% of hourly basal discrepancy but it of course depends whether another TBR is issued shortly afterwards or not etc. This mechanism is controllable via a preference option and could maybe be improved in the future._

The plugin uses the SightParser library to communicate with the [SightRemote](https://github.com/TebbeUbben/SightRemote) companion app via a secured aidl interface. Both of these are GPL3 open source projects published by @TebbeUbben

This allows AAPS to coexist with SightRemote, which is designed more as a replacement for the handset itself. This also abstracts much of the physical interfacing and separates the responsibility regarding secure access to the pump and providing information to other apps etc.

The pump has a security password which is required before access to the insulin service can be obtained. This key must be manually entered in to SightRemote via a menu item before it can deliver insulin or adjust TBRs.  When AAPS attempts to use the aidl interface for the first time SightRemote will pop-up an ALLOW/DENY security request which must also be approved.

Many thanks to everyone who has assisted to bring the project to this stage. Screenshots below, comments welcomed!

screenshot_20180131-135534
screenshot_20180131-095510
screenshot_20180130-230021


app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
```diff
@@ -140,6 +142,7 @@ private InsightPumpPlugin() {
         pumpDescription.basalMinimumRate = 0.02d;
 
         pumpDescription.isRefillingCapable = false;
```
jotomo: @jamorham Forgot to mention this earlier: this variable is oddly named: it's only `false` for MDI were priming/filling makes no sense for a pen. This flag (seach for usages) controls whether the prime/filling action is available (on the actions tab) - which I assume is supported by the Insight.

jamorham: The Insight has no remote control methods for priming cannulas or filling tubing. Does that mean the flag should stay where it is as `false`?

jotomo: Should be `true`. Priming is simply a bolus which is marked as invalid.

jamorham: Okay, I can set it as `true` to enable the priming action item, but where does the user specify the amount required to fill the cannula? On the Insight normally this is just a menu item and I don't even personally remember what the standard value is because once set during initial pump set-up I don't have to think about it anymore.

MilosKozak: right, when we merge everything together we can rename it

jotomo: @jamorham The priming dialog allows specifying amounts and you can define defaults in the preferences which the dialog will then show. You can try it with the virtual pump.

jamorham:
Okay I've set this field to `true` now

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
```diff
@@ -0,0 +1,940 @@
...
+    }
+
+    @Override
+    public boolean hasFragment() {
+        return true;
+    }
+
+    @Override
+    public boolean showInList(int type) {
+        return true;
```
jotomo: If you return `type == PUMP` here instead, the plugin still only show up as a pump plugin if you later add `ConstraintInterface` (e.g. to limit TBR to 250% or enforce a pump-specific safety limit).

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
```diff
@@ -0,0 +1,940 @@
...
+            log("Could not stop connecting - null pointer: " + e);
+        }
+    }
+
+    @Override
+    public void getPumpStatus() {
+
+        log("getPumpStatus");
+        lastDataTime = new Date();
+        if (Connector.get().isPumpConnected()) {
```
jotomo: This is always called from the queue, so there'll always be a connection open

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
```diff
@@ -0,0 +1,940 @@
...
+        }
+
+        final Cstatus cs = async.busyWaitForCommandResult(cmd, BUSY_WAIT_TIME);
+        log("Got command status: " + cs);
+    }
+
+    // Temporary Basals
+
+    @Override
+    public PumpEnactResult setTempBasalAbsolute(Double absoluteRate, Integer durationInMinutes, boolean enforceNew) {
```
jotomo: What about just converting the absolute rate to percentage and delegate to `setTempBasalPercent` to reduce code duplication?

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
```diff
@@ -0,0 +1,940 @@
...
+
+
+        fauxTBRcancel = !SP.getBoolean("insight_real_tbr_cancel", false);
+
+        final UUID cmd;
+
+        if (fauxTBRcancel) {
+            final int faux_percent = 90;
+            final int faux_duration = 15;
+            cmd = aSyncTaskRunner(new SetTBRTaskRunner(connector.getServiceConnector(), faux_percent, 15), "Faux Cancel TBR - setting " + faux_percent + "%" + " " + faux_duration + "m");
```
jotomo: Minor thing: 2nd arg should be `faux_duration` (to avoid bug when changing that var)?

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
```diff
@@ -0,0 +1,940 @@
...
+    @Override
+    public PumpEnactResult cancelTempBasal(boolean enforceNew) {
+        log("Cancel TBR");
+
+
+        fauxTBRcancel = !SP.getBoolean("insight_real_tbr_cancel", false);
+
+        final UUID cmd;
+
+        if (fauxTBRcancel) {
```
jotomo: After setting a fake neutral temp, OpenAPS might request a TBR cancellation again during the next run. With the current logic that would extend a running 90% TBR from 10m to 15m again. So you could add a check if a 90% TBR <= 15m is running and avoid the extension in that case. The Combo has implemented that logic (plus setting 110% if the cancelled TBR was > 100 to balance things out a bit).

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
```diff
@@ -0,0 +1,940 @@
...
+    }
+
+
+    private synchronized UUID deliverBolus(float bolusValue) {
+        log("DeliverBolus: " + bolusValue);
+
+        if (bolusValue == 0) return null;
+        if (bolusValue < 0) return null;
+
+        // TODO check limits here or they already occur via a previous constraint interface?
```
jotomo: General safety limits are enforced at this point, but you might need to check something like the maximum bolus the pump can deliver if you can read that value. If you want to do that, consider also implement the `ConstraintInterface`.

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
jamorham: In testing, as far as I can tell if AAPS doesn't think a TBR is set it wont try to cancel it so in this case the fake cancellation doesn't seem to get extended. I wasn't completely sure about when to set 110 vs 90 which is why I haven't implemented that logic, but I also had a thought that it might even want to be the opposite way around, apply a 110% when the temp basal had been <100% on the basis that the situation has changed and we might be trying to reverse it slightly.

AdrianLxM: Ideally AAPS would know of the 90% TBR (either directly or from reading the history). This way it would not even be real "missed" insulin as it would factor it into the IOB and would afterwards high-temp if necessary.

app/src/main/java/info/nightscout/androidaps/MainApp.java
```diff
@@ -129,6 +130,7 @@ public void onCreate() {
             if (Config.DANAR) pluginsList.add(DanaRv2Plugin.getPlugin());
             if (Config.DANAR) pluginsList.add(DanaRSPlugin.getPlugin());
             pluginsList.add(CareportalPlugin.getPlugin());
+            if (Config.APS) pluginsList.add(InsightPumpPlugin.getPlugin());
```
AdrianLxM: This is a bit non-intuitive but from times when we only had one pump: It should be `Config.DANAR` instead of `Config.APS`.
(we also should refactor that in the build config to `PUMP` or similar)

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
jotomo: @savek-cc Your logic stayed the same ;-)
@jamorham I overlooked you didn't create a `TemporaryBasal` issue or something similar. This sounds repetitive, but just delegate to `setTempBasalPercent` ;-)

app/src/main/java/info/nightscout/androidaps/MainApp.java
jamorham: I changed it as you suggested

PR #639, opened 2018-02-01
Remove jniLibs, download on build

triplem:
The jniLibs are removed and the corresponding folder is added to .gitignore. If calling "gradle clean" this path is deleted on the local file system.

Before a build the jniLibs are downloaded from github (see repositories in build.gradle) and then unzipped via the unzip task in the app-module.


MilosKozak:
working well on my laptop

PR #652, opened 2018-02-03
update bg for treatment in the past via careportal

savek-cc:
Maybe there's a better way to do it (adding a function to GlucoseStatus?) but I hope this gets the point across ;)
Basically when entering a treatment via careportal, the bg value is taken from "now" - if I change the time because I want to enter carbs from 1h ago, the bg value stays at the current one.
In order to have the right bg value for this entry in the past, I would have to look up the bg value in nightscout and then enter it here.
Using this patch, the bg value is updated with the historic value as soon as the user enters a different time or date.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/Careportal/Dialogs/NewNSTreatmentDialog.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/Careportal/Dialogs/NewNSTreatmentDialog.java
@@ -29,7 +29,6 @@ import com.wdullaer.materialdatetimepicker.date.DatePickerDialog;
 import com.wdullaer.materialdatetimepicker.time.RadialPickerLayout;
 import com.wdullaer.materialdatetimepicker.time.TimePickerDialog;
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/Careportal/Dialogs/NewNSTreatmentDialog.java
```diff
@@ -29,6 +29,7 @@
 import com.wdullaer.materialdatetimepicker.time.RadialPickerLayout;
 import com.wdullaer.materialdatetimepicker.time.TimePickerDialog;
 
+import org.apache.commons.lang3.ObjectUtils;
```
AdrianLxM: Where is this import needed?

savek-cc: Intermediate version - removed

RadoslavR:
always shows in mg / dl
screenshot_2018-02-05_221711

PR #654, opened 2018-02-03
SMB debug notifications

jotomo:
Notifications when SMBs are blocked in the background due to safety restrictions.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/Config.java
+++ b/app/src/main/java/info/nightscout/androidaps/Config.java
@@ -25,6 +25,7 @@ public class Config {
 
 
     public static final boolean displayDeviationSlope = true;
+    public static final boolean smbDebug = true;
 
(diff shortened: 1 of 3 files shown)


AdrianLxM:
When does this actually happen?
The default case is, that determine-basal will not issue an SMB if there has been a treatment within the last 3 minutes.
Only when there already was a bolus in the queue that determine-basal did not know about?

jotomo:
Yes, it might occur if one sits in the queue and the pump has connection issues and thus it takes a while before the bolus is actually issued (or the queue has just finished an SMB that took long). The checks are redundant to catch bugs. The notifications are to check (during development) if any other situations occur we didn't think of, and without the notification might have overlooked.

jotomo:
Since SMBs are blocked for **any kind** of bolus, I'm seeing these notifications after half the SMBs that follow a manual bolus.

PR #658, opened 2018-02-03
RO translation

bogdangorescu:
Added RO translation


AdrianLxM:
Hi @bogdangorescu thank you very much!
Could you please remove all lines that are `translatable="false"`
Example:
```
    <string name="key_danar_useextended" translatable="false">danar_useextended</string>
    <string name="key_danarprofile_dia" translatable="false">danarprofile_dia</string>
```
Thanks!

MilosKozak:
thanks

PR #659, opened 2018-02-03
Graph target line

jotomo:
Should the visibility of the line be configurable? I will add a pref page for home/overview when adding the buttons, I could add a toggle for this then too.


MilosKozak:
in range is not configurable as well

PR #666, opened 2018-02-05
Don't enable Insight by default

AdrianLxM:
@jamorham, this should make the Insight only active when selected in the config builder and have the Virtual Pump as default.


jamorham:
Ooops, I think I used the virtual pump as a template and was testing insight so didn't notice that.

PR #668, opened 2018-02-05
fix display unit when updating careportal entry for historic bg value

savek-cc:


savek-cc:
@AdrianLxM

AdrianLxM:
Thanks a lot!

PR #685, opened 2018-02-12
NotificationStore: add synchronisation when accessing internal store.

jotomo:
Avoids race conditions we saw reported by Crashlytics, triggered by `info.nightscout.androidaps.plugins.ConfigBuilder.ConfigBuilderPlugin#getProfile(long)` firing multiple notifications in rapid succession.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/Overview/notifications/NotificationStore.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/Overview/notifications/NotificationStore.java
@@ -21,8 +21,6 @@ import java.util.List;
 import info.nightscout.androidaps.MainApp;
 import info.nightscout.androidaps.R;
 import info.nightscout.androidaps.Services.AlarmSoundService;
-import info.nightscout.androidaps.plugins.Wear.WearPlugin;
-//Added by Rumen for snooze time 
 import info.nightscout.utils.SP;
 
 /**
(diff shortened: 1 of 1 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/Overview/notifications/NotificationStore.java
```diff
@@ -44,11 +44,11 @@ public int compare(Notification o1, Notification o2) {
         }
     }
 
-    public Notification get(int index) {
+    public synchronized Notification get(int index) {
```
MilosKozak: is it not a problem when one synchronized function is called by other?

jotomo: Sloppy work indeed ;-)
It can lead to deadlocks if one isn't careful. 
The `get` method was public, so I thought that might be a good idea, but it's only accessed internally. Furthermore, the method only does `store.get(i)` and is called in places where the store is also accessed directly as well, e.g. `store.size`. I think it's clearer what is done if the `get` method is inlined, the indirection here makes it harder to see what is going on and barely safes code.
I've added a commit that does the suggest improvements.

AdrianLxM: It should not be a problem if both functions synchronize on the same object. (In general `this`). In this case the lock is already held.

PR #687, opened 2018-02-12
Create NS-anouncements for error dialogs and local alerts.

jotomo:
screenshot_1518446323
screenshot_1518446330

Only tested in offline mode, without syncing to Nightscout server.


jotomo:
- [ ] Add preference to toggle this behaviour

PR #697, opened 2018-02-18
AndroidAPS should support more pumps

jotomo:


AdrianLxM:
Totally agreed! It should!

Issue #646, opened 2018-02-02
SMB TODO - (2.0 only wiki)


AdrianLxM:
* Check that SMB don't re-trigger a loop cycle
* Check IOBpredict / guardBG
* Move bolussnooze_dia_divisor to AMA

To wiki:
* lower autosens back to 1.2
* changed meaning of IOB

AdrianLxM:
Alarm for Carbs-Required?

jotomo:
### Bolus cancellation

**jotomo**: if an SMB is running in the background, a new bolus is requested and then attempted to cancel: this will likely cancel the SMB and then run the bolus.
**adrian**: How  about solving this on a higher level upstream? This most likely will  affect the other pump models that read history (Dana Rv2, Dana RS) or  probably even all pumps to some extent. My idea would be: on a bolus cancel always clear the CommandQueue of any bolus types (SMB or standard).  
* If a standard bolus is delivered and cancelled, an SMB should not  follow anyways as a new evaluation of the situation will be done  afterwards with the new IOB reality.
* If an SMB is delivered and cancelled (no problem if one is missing  as they are micro), the user actually would want to cancel the manually  triggered one -> delete it from the queue.

jotomo:
> Alarm for Carbs-Required?

Will require _bolussnooze_-like logic since `carbsRequired` frequently skyrockets after a meal, before carbs kick in, might take a few iterations to go down after carb intake as well.

jotomo:
- [ ] GL 364 handle deviation outliers
- [ ] GL 367 COB issues

jotomo:
- [ ] In `ConfigBuilderPlugin.applyAPSRequest`, a TBR is not set again if it's running for at least another 5m. With SMB this can lead to too much delivered insulin, if the basal insulin for the next 2h is delivered as SMB and relies on a 0% TBR to take that insulin away. Setting a TBR each and every time, to extend e.g. from 115m to 120m. every 5m is probably overkill, but maybe round to 15m or 30m, e.g. when 'rounding' to 15m, don't apply a request for 120m 0% if remaining TBR duration is 115m or 110m, but extend it when it reaches 105, thereby extending the TBR by 15m.
P.S. oref1 0.6.1 will reduce the max duration of a 0% TBR to 60m if I'm not mistaken.

McHoffi:
- [ ] Change "Notify on SMB" in Wear settings default value to true, so you do "feel" it especially in the beginning when there is given a SMB

[...]

PR #702, opened 2018-02-19
enable SMB notify by default

McHoffi:
TODO from Issue #646


MilosKozak:
@AdrianLxM ok?

AdrianLxM:
Having SMB vibrate by default is a good idea for new users or users newly using this feature.

PR #704, opened 2018-02-20
WF Layout optimations for small displays

swissalpine:
I had problems using the watchfaces nochart and bigchart on my Polar M600 because they are to big for the screen resolution of 240 * 240:
polar - unmogliche anzeige
After the optimations the watchface looks:
nochart nach aenderung
So I optimized two of the watchfaces which can used on small displays (the other aren't usable).

(I renew the PR because the file names of the screenshots were incorrect.)


AdrianLxM:
@Heiner1, could you please test the "NoChart" watchface, if it is good for you like this?

@swissalpine, updating the previews is good. There seems to have been a mixup as the test is above the chart while on the watch it should be below. (Personal remix?)

swissalpine:
@AdrianLxM No, that shouldn't be a "personal remix", for me it's an improvement for all users: I find it's faster to read if I don't need to look at the values above the graph and search after that the others below. My screen is only 34 seconds on. In this time I could read and see all informations (including graph) when the status line is above the graph. When the status line is below I often had to reactivate the screen.

AdrianLxM:
@swissalpine I did write you privately on gitter. There would be problems with round watches or square watches with higher resolution (Standard is now 320 pixels; your M600 only has 240). I believe we should adapt depending on the screen resolution and load different layouts.

Could you please check if this works: https://github.com/MilosKozak/AndroidAPS/commit/6a1b777bc3ecdeabf365ed468d4b20378d8577db

PR #705, opened 2018-02-21
Updated Insight plugin

jamorham:
This PR updates the Insight plugin to the latest version which has a connection keep-alive feature with preemptive connection and connection statistics tracking.

To facilitate this I have added a generic new event type `EventFeatureRunning` which has an enum for `Feature` this would be called for example when the bolus wizard screen opens:

```
 MainApp.bus().post(new EventFeatureRunning(EventFeatureRunning.Feature.WIZARD));
```

Subscribers to this event can take action based on what the user is doing. 

The Insight plugin responds to this event by establishing a connection to the pump *before* it is requested. When the user activates the bolus wizard this is triggered with a keep-alive timeout. 

The practical result is that the bolus request can be completed instantly as the connection is already open.  This smooths the user experience.

The effect on battery life is likely to be negligible, but there is also a preference switch to disable the behavior should the user wish to do so.

Connection statistics tracking gives a best-effort time/percentage breakdown of the time the bluetooth connection to the pump spent in various states, eg Connected, Connecting, Disconnected. During tests I saw typically around 90% disconnected.

This PR also includes a change to the AndroidAPS `QueueThread` wake-lock changing it from: `SCREEN_DIM_WAKE_LOCK` to `PARTIAL_WAKE_LOCK` 

I don't believe there is any advantage to keeping the screen on to complete network/bluetooth operations and the construction of the existing wake-lock I don't think it would light the screen when it was already off and this type of wake-lock is released if the user does lock the screen anyway.


app/src/main/java/info/nightscout/androidaps/queue/QueueThread.java
```diff
@@ -39,7 +39,8 @@ public QueueThread(CommandQueue queue) {
 
         this.queue = queue;
         PowerManager powerManager = (PowerManager) MainApp.instance().getApplicationContext().getSystemService(Context.POWER_SERVICE);
-        mWakeLock = powerManager.newWakeLock(PowerManager.SCREEN_DIM_WAKE_LOCK, "QueueThread");
+        //mWakeLock = powerManager.newWakeLock(PowerManager.SCREEN_DIM_WAKE_LOCK, "QueueThread");
```
jotomo: What's the value of keeping this as a comment?

jamorham: none other than as an illustration/reminder as to the previous behavior

jamorham:
thanks

app/src/main/java/info/nightscout/androidaps/queue/QueueThread.java
jotomo: Just my personal opinion: without a comment stating why the previous behaviour was in place or why it was changed (which I think would be better placed in the commit message instead), this just pollutes the source code, since by your argument you could just comment out all code and never remove any. As a reader of that code (who didn't write it), this only says "something else can be done here as well", which is frankly pretty useless to me, I'd still have to figure out Adrian wrote the initial code (by annotating an earlier version of the file) and ask him why he did so.

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
```diff
@@ -470,6 +471,22 @@ public PumpEnactResult setTempBasalAbsolute(Double absoluteRate, Integer duratio
...
+        // TODO this is experimental and not enabled due to preference option
+        // ignore TBR changes if the setting is enabled and they are not 0% but are +- 30% of where we are now or we've been running is TBR < 15 minutes
+        if ((percent_amount != 0)
+                && (SP.getBoolean("insight_dont_change_active_tbr", false)
+                && MainApp.getConfigBuilder().isTempBasalInProgress())) {
+            final TemporaryBasal temporaryBasalFromHistory = MainApp.getConfigBuilder().getTempBasalFromHistory(System.currentTimeMillis());
+            if ((temporaryBasalFromHistory != null) && (temporaryBasalFromHistory.isInProgress())) {
+                // only change if it is +- 30% from where we are now or we have done more than 15 minutes already
+                if ((Math.abs(temporaryBasalFromHistory.percentRate - percent_amount) <= 30)
+                        || (temporaryBasalFromHistory.getRealDuration() < 15)) {
```
AdrianLxM: Shouldn't this be `&&  (temporaryBasalFromHistory.getRemaining() < 15` ?

In my opinion it definitely should change, if only a few minutes have passed meaning a lot of the rate is still remaining -> the difference in insulin will be large if not changed!

This looks like a very dirty hack where part of the algorithm is put into the driver and IMHO should be cleaned up.

PR #708, opened 2018-02-21
Watch: don't bother pump with carbs if pump doesn't support it.

jotomo:


MilosKozak:
@AdrianLxM ok?

AdrianLxM:
Yes, it's ok :)

Issue #696, opened 2018-02-18
Combo clean up tasks


jotomo:
- [x] Remove CommandQueue.setProfile hack for initial profile switch
- [x] Properly fix QueueCommandTest.
- [x] (maybe) Integrate fragment history actions with queue and enable them
- [x] Switch Answer calls to FabricPrivacy.
- [x] Add to pumpdescription: `pumpDescription.tempDurationStep15mAllowed = true;
        pumpDescription.tempDurationStep30mAllowed = true;`
- [x] Produce a _Treatment_ object in _ComboPlugin.deliverTreatment_, when dispatching _EventOverviewBolusProgress_ (_isSMB_ flag is needed to suppress notifications on watch). #732 
- [x] #715

MilosKozak:
@jotomo what is the last task?

jotomo:
@MilosKozak Last open point is  about reading TDD and error history. TDD is now done via _CommandLoadTDDs_ thans to Adrian, but error history is still open. Currently, the Combo foregoes the queue to read that. I was pondering adding a generic _CommandCustom_ that would allow to run an arbitrary callback as part of the queue. That would allow pump drivers to run driver-specific stuff cleanly ...

PR #709, opened 2018-02-22
Add Fabric Privacy option

jamorham:
Fabric Answers and Crashlytics provide an excellent service for developers but this also is a risk to the user's privacy. Although developers receive mostly anonymized data from the fabric console, fabric itself receives much more information.  Fabric is now owned by Google.

With any data upload which can potentially contain private information, such as app usage data etc, it is my belief that the user should be given a choice to opt-out of this data collection if they wish to do so. This is consistent with electronic data processing regulations present in the EU.

Fabric doesn't make this as easy is they could and so I have created a wrapper for both `Answers` and `Crashlytics` which respects a new data choices advanced preference option.

In the changes it should be obvious that wherever either of these features are used they now access the wrapper which facilitates the opt out. 

Because Crashlytics receives data on initialization, the initialization process also is conditional on this data choice being set. The default option is enabled - the app behavior is only modified if the user explicitly disables the data collection upload. 

screenshot_20180222-123503


MilosKozak:
@jotomo are you able to cherrypick and adapt combo branch too?

jotomo:
@MilosKozak That would change the outcome of our bet ;-) It's an easy search & replace I'll do after the merge to dev. I've added an item for it to https://github.com/MilosKozak/AndroidAPS/issues/696

jamorham:
@jotomo it is just a search and replace other than the set up code changes in `MainApp`

@MilosKozak thanks

PR #710, opened 2018-02-22
Quick fix for Insight event handling when not enabled

jamorham:
This is a quick fix to prevent the crash if the event handler gets fired when not using Insight. I will update the lifecycle for the connector so it only is instantiated when the plugin is enabled rather than when it is instantiated but for now this will fix the crash.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/connector/Connector.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/connector/Connector.java
@@ -494,16 +494,18 @@ public class Connector {
 
     @Subscribe
     public void onStatusEvent(final EventFeatureRunning ev) {
-        if (SP.getBoolean("insight_preemptive_connect", true)) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/connector/Connector.java
```diff
@@ -494,16 +494,18 @@ private long getEntryTime(Map.Entry entry) {
...
-            switch (ev.getFeature()) {
-                case WIZARD:
-                    log("Wizard feature detected, preconnecting to pump");
-                    connectToPump(120 * 1000);
-                    break;
-                case MAIN:
-                    log("Main feature detected, preconnecting to pump");
-                    connectToPump(30 * 1000);
-                    break;
+        if (isConnected()) {
```
MilosKozak: `isConnected` ???
I didn't test it but would not be correct test if plugin is enabled?

PR #723, opened 2018-03-01
Not everybody's profile & targets are in mg/dl :)

PoweRGbg:


app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
```diff
@@ -69,7 +70,7 @@
     private static final double FAV1_DEFAULT = 5;
     private static final double FAV2_DEFAULT = 10;
     private static final double FAV3_DEFAULT = 20;
-
+    final double tempTargetValue = 140d;
```
AdrianLxM: This constant seems to be unused.

PoweRGbg: removed

PoweRGbg:
Could anyone explain why these codecov checks fail ?!?

jotomo:
The new insulin button already has a checkbox to start an eating soon TT (and allows giving a pre-bolus at the same time).
That dialog needs to be updated to deal with mmol properly as well. Code to do so should use the `Constants` class and the methods in `Profile` to do calculations. The formatting code would best be placed in a utility class rather than duplicate the code. I was thinking about extracting the formatting code altogether to make it cleaner and (including checking if any option is set).

app/src/main/java/info/nightscout/androidaps/MainApp.java
```diff
@@ -139,7 +139,7 @@ public void onCreate() {
             if (Config.DANAR) pluginsList.add(DanaRv2Plugin.getPlugin());
             if (Config.DANAR) pluginsList.add(DanaRSPlugin.getPlugin());
             pluginsList.add(CareportalPlugin.getPlugin());
-            if (Config.DANAR) pluginsList.add(InsightPumpPlugin.getPlugin());
+            // if (Config.DANAR) pluginsList.add(InsightPumpPlugin.getPlugin()); // <-- Enable Insight plugin here
```
jotomo: Why are you manually copying over changes from `dev` instead of rebasing/merging? This obfuscates the original change and makes it harder to track why it was changed and who originally changed it.

PoweRGbg: Because I don't have the ram for 10 instances of android studio

PoweRGbg: I fix for the Insulin button is also done

PoweRGbg: @jotomo Consider Eating soon mode in carbs button a mealAnnouncement feature, that I use often. And the other suggestions are just a fix for me to be able to use these buttons. If you don't like the way I did it - please fix it your way.

jotomo: * Sorry I overlooked the _mmol_ changes to the insulin button.
* You can take over changes from `dev` using Ctrl-T, no need for 10 instances of AS.
* Logically, placing the "eating soon TT" in the insulin dialog made sense for me, since I combine it frequently with a bolus and "eating soon TT" is about getting insulin into system in preparation for a meal, not the meal itself.

jotomo: > And the other suggestions are just a fix for me to be able to use these buttons. If you don't like the way I did it - please fix it your way.

I'm concerned about making the code maintainable, not doing it any one's way. Using Constants prevents error from typos in strings, e.g. using `Constants.MMOL` and using methods like `Profile.toMmol` allows having conversion code in one place.

PoweRGbg: @jotomo Does these changes represent what you ment ? ( using Profile.toMmol method and Constants ) 
If it is ok now I should figure a way to remove the the changes in MainApp.java

MilosKozak:
checking one checkbox should exclude the other. this can be confusing when you check both

PoweRGbg:
@MilosKozak I do a check and if both are selected - the bigger ( higher ) temp is applied, the other way is to do it with Drop down list, but it will not be so good looking as Johanes did it

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
```diff
@@ -235,80 +237,121 @@ private void submit() {
...
             double prefTT = SP.getDouble(R.string.key_activity_target, 140d);
-            double tt = prefTT > 0 ? prefTT : 140d;
-            if (startActivityTTCheckbox.isChecked()) {
-                confirmMessage += "<br/>" + "TT: " + "<font color='" + MainApp.sResources.getColor(R.color.high) + "'>" + ((int) tt) + "mg/dl for " + ((int) ttDuration) + " min </font>";
-            }
 
+            double tt = 140d;
+            double esTT = 90d;
+            Profile currentProfile = MainApp.getConfigBuilder().getProfile();
+            if(currentProfile.equals(null))
```
jotomo: This will fail if _currentProfile_ is actually null. You probably meant `currentprofile == null` :-)
Not sure what to do in that situation. The dialog will just close without any action?

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewInsulinDialog.java
```diff
@@ -246,8 +246,19 @@ private void submit() {
             double ttDuration = prefTTDuration > 0 ? prefTTDuration : 45d;
             double prefTT = SP.getDouble(R.string.key_eatingsoon_target, 80d);
             double tt = prefTT > 0 ? prefTT : 80d;
+            Profile currentProfile = MainApp.getConfigBuilder().getProfile();
+            if(currentProfile.equals(null))
```
jotomo: currentProfile == null

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewInsulinDialog.java
```diff
@@ -246,8 +246,19 @@ private void submit() {
...
+                return;
+            if(currentProfile.getUnits().equals(Constants.MMOL)) {
+                tt = prefTT > 0  ? Profile.toMgdl(prefTT, Constants.MGDL) : 80d;
+            } else
+            tt = prefTT > 0  ? prefTT : 80d;
```
jotomo: The _if_-branch has brackets around it, the _else_-branch not, which is also not aligned. This makes it hard to read. In general, I try to be consistent: either use no brackets or brackets for both _if_ and _then_.

app/src/main/java/info/nightscout/androidaps/MainApp.java
jotomo: Yes, that's what I meant. Thanks for updating. And thanks for catching the missing conversion. I remembered to make everything configurable, but forget about _mmol_ users (again) ;-)

jotomo:
@PoweRGbg when you announce a meal with "eating soon TT", that's the only action you take, right? Functionally, that's easily doable with the insulin button. I still like it there to combine with a pre-bolus. That also avoids the issue of which TT to use if both are selected ;-)

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewInsulinDialog.java
PoweRGbg: Thanks - fixed in the latest commit

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewInsulinDialog.java
PoweRGbg: Thanks - fixed in the latest commit

PoweRGbg:
@jotomo I usually do an low TT for 120 min with the meal start, that helps SMB to be a little more aggressive , and deals better with fatty meals. I like that option to be in the carbs also, not to go to TT and set it manually

MilosKozak:
I'd keep checkboxes. just add hook and unselect the second when first is checked. I agree checkboxes are nicer.
Current way you can select both but only one is executed

jotomo:
> I usually do an low TT for 120 min with the meal start, that helps SMB to be a little more aggressive , and deals better with fatty meals. I like that option to be in the carbs also, not to go to TT and set it manually

@PoweRGbg I do the same :-) Just to be extra sure we're not misunderstanding each other: exactly that functionality is present from the _insulin_ button, with presets for an eating soon TT, without manually entering it.

PoweRGbg:
@MilosKozak With the latest commit only one of the two TT checkboxes can be selected
@jotomo My case is: I usually enter carbs > 35, bolusInsulin = 0, tempTargetType = "Eating soon", tempTargetDuration >120 min . In this case in the current dev branch I need to press Insulin, check "Start eating soon TT" and confirm, press Carbs, enter carbs and confirm. But with the propoosed changes I can do that with 1 step less.

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
```diff
@@ -69,9 +71,9 @@
     private static final double FAV1_DEFAULT = 5;
     private static final double FAV2_DEFAULT = 10;
     private static final double FAV3_DEFAULT = 20;
-
     private CheckBox suspendLoopCheckbox;
     private CheckBox startActivityTTCheckbox;
+    private CheckBox ESMCheckbox;
```
AdrianLxM: Could you please rename this to start wit a lowercase letter.

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
```diff
@@ -235,80 +249,122 @@ private void submit() {
             }
 
             double prefTTDuration = SP.getDouble(R.string.key_activity_duration, 90d);
```
AdrianLxM: As there are now multiple TTs, could you please rename this to something like `activityTT...`
(same with the duration that should be sorted to be next to this line imho)

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
```diff
@@ -235,80 +249,122 @@ private void submit() {
...
+
+            double tt = 140d;
+            double esTT = 90d;
+            Profile currentProfile = MainApp.getConfigBuilder().getProfile();
+            if(currentProfile == null)
+                return;
+            if(currentProfile.getUnits().equals(Constants.MMOL)) {
+                esTT = eatingSoonTT > 0  ? eatingSoonTT*Constants.MMOLL_TO_MGDL : 90d;
+                tt = prefTT > 0  ? prefTT*Constants.MMOLL_TO_MGDL : 140d;
+            } else
```
AdrianLxM: This is a dangling else? Only affects one line but from the indentation it looks like it should it should actually inclute the line `tt = prefTT > 0  ? prefTT : 140d;` as well?

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
```diff
@@ -235,80 +249,122 @@ private void submit() {
...
             double prefTT = SP.getDouble(R.string.key_activity_target, 140d);
-            double tt = prefTT > 0 ? prefTT : 140d;
+
+            double tt = 140d;
+            double esTT = 90d;
+            Profile currentProfile = MainApp.getConfigBuilder().getProfile();
+            if(currentProfile == null)
+                return;
+            if(currentProfile.getUnits().equals(Constants.MMOL)) {
+                esTT = eatingSoonTT > 0  ? eatingSoonTT*Constants.MMOLL_TO_MGDL : 90d;
```
AdrianLxM: We do have a function for this in the Profile class. (`Profile#toMgdl`)

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
PoweRGbg: ok ?

PoweRGbg:
@AdrianLxM all fixed  ?

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewInsulinDialog.java
```diff
@@ -246,8 +246,19 @@ private void submit() {
             double ttDuration = prefTTDuration > 0 ? prefTTDuration : 45d;
             double prefTT = SP.getDouble(R.string.key_eatingsoon_target, 80d);
             double tt = prefTT > 0 ? prefTT : 80d;
+            Profile currentProfile = MainApp.getConfigBuilder().getProfile();
+            if(currentProfile == null)
+                return;
+            if(currentProfile.getUnits().equals(Constants.MMOL))
+                tt = prefTT > 0  ? Profile.toMgdl(prefTT, Constants.MGDL) : 80d;
```
AdrianLxM: This must be `tt = prefTT > 0  ? Profile.toMgdl(prefTT, Constants.MMOL) : 80d;` as we are in the MMOL branch.

It should be possible to reduce the whole if-else construct by something like

`tt = prefTT > 0  ? Profile.toMgdl(prefTT, currentProfile.getUnits()) : 80d;`

MilosKozak:
finished by https://github.com/MilosKozak/AndroidAPS/commit/ec1c43b905f17468d1c8035ed834a676685c2745

RadoslavR:
screenshot_20180310-221152
screenshot_20180310-221300

PR #732, opened 2018-03-03
Combo: Reenable Disconnet for 15 min / 30 min

swissalpine:
Consequence of Milos' commit c61fcdb is to reallow disconnect for 15/30 minutes for the combo which is capable to do so.


swissalpine:
Hm. This PR seems to be for later, because the changes in dev branch didn't concerne this branch or the master branch. In my Combo-Dev-Mix this works fine and reenables the Disconnect Feature.

PR #743, opened 2018-03-09
removed local settings and for NS staleData

PoweRGbg:


PoweRGbg:
NS staleData alarm will ring only if OPENAPS_ENABLE_ALERTS is TRUE in NS site enviroment

MilosKozak:
@AdrianLxM ?

AdrianLxM:
This PR would need a rewrite of `isAlarmForStaleData()` in order to remove the offline alert functionallity that already exists in another plugin.
If I see it correctly the offline functionality wouldn't be triggered anyway as it can only be triggered from the NSClient.

close?

PR #745, opened 2018-03-10
Update to SightRemote 1.4.0

TebbeUbben:
This updates AndroidAPS for SightRemote 1.4.0 (compatibility version: Asclepenius).

It also adds the ability to change the basal profile using AndroidAPS. It won't trigger an update if the difference between the pump and the profile is only 0.02 or less. This has to do with some rounding errors which would cause an update on every connection otherwise.


MilosKozak:
@jamorham @TebbeUbben it's on you. I don't have insight to test. tell me if you want to merge it as it is

PR #747, opened 2018-03-11
Removed unneeded conversion to Md/dl which leads to a mgDl targets

PoweRGbg:
even if the profile is in MMOL


MilosKozak:
I forgot this. thanks!

PR #753, opened 2018-03-11
BaseWatchFace.java: Update chart with bolus and prediction data.

andrew-warrington:
Copying across innovations from AdrianLxM on the BigChart to other watch faces. Affects AAPSv2 and Steampunk.


AdrianLxM:
Thanks!

PR #755, opened 2018-03-11
Show bolus progress

TebbeUbben:


app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPumpPlugin.java
```diff
@@ -497,7 +497,10 @@ public PumpEnactResult deliverTreatment(DetailedBolusInfo detailedBolusInfo) {
         if (result.success) {
             log("Success!");
 
+            Treatment t = new Treatment();
+            t.isSMB = detailedBolusInfo.isSMB;
             final EventOverviewBolusProgress bolusingEvent = EventOverviewBolusProgress.getInstance();
+            bolusingEvent.t = t;
```
jotomo: @AdrianLxM I missed this for the Combo (assigning _Treatment_), which breaks the _Disable vibrations on watch for SMB_. I was thinking of either creating a method to convert a _DetailedBolusInfo_ to _Treatment_ or see if the field can be changed to _DetailedBolusInfo_, so that when other fields will be accessed, we won't run into issues with a partially filled _Treatment_ instance. Or replace with parameters we of stuff we actually use, _insulin_ and _SMB_ as far as I've seen from a quick lock. Actually, _insulin_ is only written to, the only time the _Treatment_ instance is used, is to check the SMB flag. Progress is reported via the _progress_ field. So we could just replace the _Treatment_ instance  with an _isSMB_ flag.

jotomo: I've added it to the SMB TODO, since this isn't about this PR.

PR #760, opened 2018-03-13
(Re)enable test coverage in travis.yml.

jotomo:


MilosKozak:
better solution found . thx

PR #765, opened 2018-03-15
Draw basals in Profile Viewer

AdrianLxM:
screenshot_20180315-135856


MilosKozak:
nice!!!

MilosKozak:
would be possible to adapt DanaX profile viewer to use this class too?

AdrianLxM:
> would be possible to adapt DanaX profile viewer to use this class too?

done
screenshot_20180315-165539

PR #766, opened 2018-03-16
Chart menu

AdrianLxM:


MilosKozak:
remove unused strings?

AdrianLxM:
The Strings are still used in the History-Browser

MilosKozak:
let's change it there too

AdrianLxM:
I thought about that, but in the History-Browser we have no problem with space (no current pump/loop state, no action buttons) and the logic is different: It does not store to the main settings?!

If you want to, I can change it there too, up to you.

MilosKozak:
ahhh ok, keep it as it is now

jotomo:
COB & Deviations used to disable each other upon activation. That doesn't happen anymore with the dialog. Both enabled at the same time makes for a barely usable graph, so I'd say it makes sense to keep that behaviour.

AdrianLxM:
Same PR but with history browser: https://github.com/MilosKozak/AndroidAPS/pull/768 I prefer the History-Browser as is, but have no strong opinion.

AdrianLxM:
screenshot_20180317-000332

@jotomo having everything enabled is too much, I agree. But the mutual exclusion of COB and dev seemd to be quite arbitrary to me. Especially as the dymanic cob calculation depends on DEV i like to have a look at the overlay from time to time.

jotomo:
@AdrianLxM Point taken. Lets leave it up to the user what to display and what not.

jotomo:
LGTM; works and code looks good

AdrianLxM:
Thanks for testing @jotomo.

> ahhh ok, keep it as it is now
@MilosKozak sorry, I overread this while already preparing the changes at the history-browser. Mergin without the changes now.

PR #767, opened 2018-03-16
Don't overlap time ticks with current time

AdrianLxM:


jotomo:
@AdrianLxM Thanks, I wanted to say yesterday that this has been working well for a week now, but then I forgot :-)

PR #771, opened 2018-03-17
Next round of button updates

jotomo:
* Broken calculation of pre-bolus removed
* When nothing is selected, inform the user that nothing will happen
* Carbs button: option "Suspend loop 30m" (to prevent the loop from overcorrecting hypo carbs), replaced with a configurable "hypo TT" to better control the hypo outro.


jotomo:
Hi @PoweRGbg I've updated the carbs button. I'm hoping I got the mmol part right this time. Would be cool if you could give it a test! :-)
screenshot_20180318-132026

MilosKozak:
👍

PoweRGbg:
Sometimes when using the carbs button - AAPS connects to the pump - why ?

AdrianLxM:
@PoweRGbg sometimes or always?
I'm currently testing the Combo setup and the Combo can't sotre carbs.
The Dana*RS (actually the Dana*R also) can store treatments. With the RS this is really nice as you can switch phones and continue looping while offline as the treatment data will be read from the pump.

PR #773, opened 2018-03-17
 allow profile to be null

MilosKozak:


app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
```diff
@@ -187,7 +187,9 @@
 
     final Object updateSync = new Object();
 
-    public enum CHARTTYPE {PRE,BAS, IOB, COB, DEV, SEN};
+    public enum CHARTTYPE {PRE, BAS, IOB, COB, DEV, SEN}
+
+    ;
```
AdrianLxM: ... here the semicolon wanted to run away from the enum ;)

AdrianLxM:
From first code review: looks good! :)

As there is no fallback to the profile anymore, it can be removed to write the basals to the pump when a new profilestore comes in: https://github.com/MilosKozak/AndroidAPS/blob/dev/app/src/main/java/info/nightscout/androidaps/plugins/ProfileNS/NSProfilePlugin.java#L124

Also in Data-Service there is this comment with following code:
```
// if there are no profile switches this should lead to profile update
if (MainApp.getConfigBuilder().getProfileSwitchesFromHistory().size() == 0)
MainApp.bus().post(new EventNewBasalProfile());
```
https://github.com/MilosKozak/AndroidAPS/blob/dev/app/src/main/java/info/nightscout/androidaps/Services/DataService.java#L371

PR #776, opened 2018-03-19
New constraints design

MilosKozak:
Add feature to provide reason for every constraint
image


stavlor:
:+1:

app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
```diff
@@ -451,16 +449,14 @@ public void applySMBRequest(APSResult request, Callback callback) {
...
-    public boolean isLoopEnabled() {
-        boolean result = true;
+    public void limitRunningLoop(BooleanConstraint value) {
 
         ArrayList<PluginBase> constraintsPlugins = MainApp.getSpecificPluginsListByInterface(ConstraintsInterface.class);
         for (PluginBase p : constraintsPlugins) {
             ConstraintsInterface constrain = (ConstraintsInterface) p;
             if (!p.isEnabled(PluginBase.CONSTRAINTS)) continue;
-            result = result && constrain.isLoopEnabled();
+            constrain.limitRunningLoop(value);
```
jotomo: Could fix the typo, constrain**t** here and in the package name? My mind trips over this every time ;-)

app/src/main/java/info/nightscout/androidaps/interfaces/constrains/Constraint.java
```diff
@@ -0,0 +1,27 @@
+package info.nightscout.androidaps.interfaces.constrains;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Created by mike on 19.03.2018.
+ */
+
+public class Constraint {
```
jotomo: If you generify this, BooleanConstraint, IntegerConstraint (...) won't be needed :-)

app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
MilosKozak: already done in next PR , thx anyway

app/src/main/java/info/nightscout/androidaps/interfaces/constrains/Constraint.java
MilosKozak: that was my 1st version :) but this is somehow cleaner

MilosKozak: i'll think about it again

jotomo: ... or just refactor **if** it still makes sense when we actually need more types and see more concrete requirements for the type.

app/src/main/java/info/nightscout/androidaps/MainApp.java
```diff
@@ -191,13 +194,14 @@ public void onCreate() {
...
         NSUpload.uploadAppStart();
+
         if (Config.NSCLIENT)
             FabricPrivacy.getInstance().logCustom(new CustomEvent("AppStart-NSClient"));
         else if (Config.G5UPLOADER)
             FabricPrivacy.getInstance().logCustom(new CustomEvent("AppStart-G5Uploader"));
         else if (Config.PUMPCONTROL)
             FabricPrivacy.getInstance().logCustom(new CustomEvent("AppStart-PumpControl"));
-        else if (MainApp.getConfigBuilder().isClosedModeEnabled())
+        else if (MainApp.getConstraintChecker().limitClosedLoop(new Constraint<>(true)).get())
```
AdrianLxM: The wording is confusing here. "limitClosedLoop(...true)" sounds like the command to limit a closed loop.

MilosKozak: what would you recommend?

AdrianLxM:
I would call it "isClosedModeEnabled" as before.

The ConstraintChecker does not necessarily have to implement the same interface as all Plugins that do imposte constraints. this way the dummy `new Constraint<>(true)` could be called within the ContraintChecker.

Should I do this in code to make it more clear what I mean?

MilosKozak:
and the interface function leave `limitClosedLoop`?

AdrianLxM:
@MilosKozak this is what I meant: https://github.com/MilosKozak/AndroidAPS/commit/01a9e7682750908bb7adf5f315eb6e810088e312

MilosKozak:
make PR

AdrianLxM:
to dev or to this branch?

MilosKozak:
this

PR #779, opened 2018-03-19
Loop results

jotomo:


app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
```diff
@@ -379,37 +378,54 @@ public void applyTBRRequest(APSResult request, Profile profile, Callback callbac
...
+        }
+
+        long lastBolusTime = getLastBolusTime();
+        if (lastBolusTime != 0 && lastBolusTime + 3 * 60 * 1000 > System.currentTimeMillis()) {
+            log.debug("SMB requested but still in 3 min interval");
+            if (callback != null) {
+                callback.result(new PumpEnactResult()
+                        .comment(MainApp.gs(R.string.smb_frequency_exceeded))
+                        .enacted(false).success(false)).run();
+            }
```
AdrianLxM: isn't there a `return` missing to guard from further execution and still adding it to the `CommandQueue`?

app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
```diff
@@ -395,6 +395,7 @@ public void applyTBRRequest(APSResult request, Profile profile, Callback callbac
             }
         } else if (activeTemp != null
                 && activeTemp.getPlannedRemainingMinutes() > 5
+                && request.duration - activeTemp.getPlannedRemainingMinutes() < 30
```
jotomo: This is an attempt to balance safety and _not setting a TBR every time_. Previously, a same-rate TBR was only applied if it was about to experience. This logic extends a running TBR if the new TBR is 30+ longer than the old. E.g. if oref1 requests a 0% 60m TBR and 20m of an existing 0% TBR are running, this will extend the TBR, so it won't overdeliver basal insulin administered via SMB already.
WDYT? 30m a good compromise? 15m?
@AdrianLxM @MilosKozak

app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
jotomo: Good you saw that, thanks! Fixed now.

app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
AdrianLxM: I think the 30 Minutes are a good compromise. With "AMA" that only sets 30 minutes it does not change anything. With SMB it really is needed to have longer low-temps as SMB is a super-bolus.

Issue #783, opened 2018-03-20
Closed loop not running


infoneo:
After commit ad3bf19e4c2857d93f59614de3d8aeb8448d78bb (after change in the /plugins/IobCobCalculator/IobCobCalculatorPlugin.java - onEventNewBG function) loop is not starting automatically (only manually). I use xDrip + G5 with replaced batteries.

jotomo:
Try to pause NSClient.
It seems the SGV from NS is processed first, but doesn't trigger the loop since it's not the active BG source, and when processing the original xDrip value next, the existing SGV is only updated, not triggering a loop run either (only new values do).
Duplicate of #782

McHoffi:
When pausing NSClient it is working normally.

jotomo:
@infoneo @McHoffi @swissalpine Fix ready for testing: https://github.com/MilosKozak/AndroidAPS/pull/786

McHoffi:
Not working for me, still not enacted.
[AndroidAPS.log](https://github.com/MilosKozak/AndroidAPS/files/1830344/AndroidAPS.log)

infoneo:
For me, the fix has resolved problem - the loop is enacted periodically.

jotomo:
@infoneo @McHoffi @swissalpine Thanks for testing! The proposed fix won't work in all scenarios. I've reverted the initial commit causing this issue (see https://github.com/MilosKozak/AndroidAPS/pull/787).

infoneo:
Thanks!

PR #786, opened 2018-03-20
Remove EventNewBg.isNew().

### LINKED ISSUE #782: Loop not enacted every 5 min
I noticed the loop was not enacted, because the prediction line was behind.
I use the Build: fd1e78ac-2018.03.19 (tdds-again branch)

screenshot_20180320-124054_androidaps

[AndroidAPS.log](https://github.com/MilosKozak/AndroidAPS/files/1829124/AndroidAPS.log)

jotomo:
If a local BG source produces a new reading that is sent to NS, the
same reading is retrieved from NS and might be processed before
the local value is. That leads to the NS value creating a new
record, but since it's not the active BG source it doesn't trigger
a loop run. When the local value is processed it doesn't trigger a
loop run either since it wasn't a value.

This commit therefore removes the isNew field. Consumers query
isFromActiveBgSource() and isCurrent() as neede, which should be
enough.

Fixes #782, #783.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
+++ b/app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
@@ -240,7 +240,7 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
             log.error("Unhandled exception", e);
         }
         VirtualPumpPlugin.setFakingStatus(true);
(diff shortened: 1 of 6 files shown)


swissalpine:
This PR solves for me the issue. AAPS loops steadily also if Nightscout Client is activated.
@infoneo @McHoffi @jotomo

PR #789, opened 2018-03-20
xdrip plugin - advancedFilteringSupported in eng mode

infoneo:
xDrip+ can act as G4 Receiver, it also calculates noise lever. Using it with caution should be allowed in eng. mode with all smb modes.


jotomo:
Xdrip does different noise processing and the reported noise levels are not yet used by AAPS. Enabling this option comes after implementing noise handling in AAPS. Before that, there's no point in enabling and testing this - the engineering mode is not meant for running the loop, but for developing such features.

infoneo:
ok.thx

PR #799, opened 2018-03-23
Prime/fill button refresh and some drive-by improvements

jotomo:
screenshot_20180323-180651

### DIFF
--- a/app/src/main/java/info/nightscout/utils/NSUpload.java
+++ b/app/src/main/java/info/nightscout/utils/NSUpload.java
@@ -16,8 +16,10 @@ import org.slf4j.LoggerFactory;
 
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 
 import info.nightscout.androidaps.MainApp;
 import info.nightscout.androidaps.R;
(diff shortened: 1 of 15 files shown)


app/src/main/java/info/nightscout/utils/NSUpload.java
```diff
@@ -169,24 +157,35 @@ public static void uploadExtendedBolusEnd(long time, long pumpId) {
...
+    }
+
+    private static synchronized String getNextAvailableSecond(long time) {
+        long second = time - time % 1000;
+        if (second == lastUsedSecond) second += 1000;
+        lastUsedSecond = second;
+        return DateUtil.toISOString(second);
+    }
+
     public static void uploadDeviceStatus() {
```
jotomo: ... I'm implying all uploads are new records. I'm not sure if this holds true for `uploadDeviceStatus` and `updateProfileSwitch`, in which cases it is desired to use the same timestamp so an existing record is updated.

jotomo: @MilosKozak ... I'm implying all uploads are new records. I'm not sure if this holds true for `uploadDeviceStatus` and `updateProfileSwitch`, in which cases it is desired to use the same timestamp so an existing record is updated.

MilosKozak:
finished?

app/src/main/java/info/nightscout/utils/NSUpload.java
```diff
@@ -204,11 +213,11 @@ public static void uploadDeviceStatus() {
             if (lastRun != null && lastRun.lastAPSRun.getTime() > System.currentTimeMillis() - 300 * 1000L) {
                 // do not send if result is older than 1 min
                 APSResult apsResult = lastRun.request;
-                apsResult.json().put("timestamp", DateUtil.toISOString(lastRun.lastAPSRun));
+                apsResult.json().put("timestamp", getNextAvailableSecond(lastRun.lastAPSRun.getTime()));
```
jotomo: @MilosKozak Should be done. I think I accidentally deleted a question: in NSUpload I'm now checking that when the timestamp used is unique and adding a second to make it unique when needed. This assumes all uploads create new entries, rather than updating existing entries (where the timestamp needs to match, right?). I'm not sure if my assumptions hold on this method (uploadDeviceStatus) and uploadProfileSwitch...

MilosKozak: deduping is done on NS side

MilosKozak: if necessary

app/src/main/java/info/nightscout/utils/NSUpload.java
```diff
@@ -337,7 +337,7 @@ public static void uploadCareportalEntryToNS(JSONObject data) {
                 if (data.has("notes")) prebolus.put("notes", data.get("notes"));
                 long mills = DateUtil.fromISODateString(data.getString("created_at")).getTime();
                 Date preBolusDate = new Date(mills + data.getInt("preBolus") * 60000L + 1000L);
-                prebolus.put("created_at", getNextAvailableSecond(preBolusDate.getTime()));
+                prebolus.put("created_at", DateUtil.toISOString(preBolusDate.getTime()));
```
MilosKozak: why here too?

MilosKozak: aaa right it's used for syncing

jotomo: Just testing whether you're properly reviewing my stuff ;-)
... meant to fix the two _updateProfileSwitch_ and screwed up but ended up in the wrong method. Next fix pushed.

jotomo: ... which one is correct now? :-)

PR #813, opened 2018-03-28
Clean up initial profile push to pump.

jotomo:


jotomo:
Tested with the Combo.

PR #815, opened 2018-03-29
Cleanup imports, rename method

triplem:
Minor "refactoring". Rename a method and cleanup imports.


AdrianLxM:
Did you want to add those files?

 app/full/release/output.json
 app/release/output.json

triplem:
Removed files and provide new PR.

PR #819, opened 2018-04-01
Unbreak wear plugin

jotomo:


app/src/main/java/info/nightscout/androidaps/HistoryBrowseActivity.java
```diff
@@ -233,7 +233,7 @@ void updateGUI(String from) {
 
         //  ------------------ 1st graph
 
-        final GraphData graphData = new GraphData(bgGraph, IobCobCalculatorPlugin.getPlugin());
```
MilosKozak: it must be here. History browser is not finished yet. it's going to be necessary use another instance of Iob calculator because it will hold different time range

PR #821, opened 2018-04-01
Gradle Update to 4.4

triplem:


MilosKozak:
looks good. thanks.
what about the exception at the end? ignore?

MilosKozak:
sorry looks like my bug

PR #832, opened 2018-04-03
LocalProfile reset/save button

AdrianLxM:
Instead of saving the local profile on every tiny edit it will now replace the "Activate Profile" button with a "reset" and "save" button when there had been changes.

 * The "save" button will not show, if the currently edited profile is not vaild.
 * Pressing "save" will persist the profile as before
 * Pressing "reset" will reload the data from the settings and rebuild the view elements.
 * "Activate Profile" is only visible if the data was not edited (when freshly loaded or after "reset" or "save")
 * "Invalid Profile" is calculated based on the edited state, not the stored state.

As it does not auto-save any more, I hope that this solves the issues with scrambled profiles when the fragment's view gets destroyed while updating etc.


jotomo:
Tested and working

PR #839, opened 2018-04-04
Remove superfluous update in onResume.

jotomo:


app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
```diff
@@ -841,14 +836,6 @@ public void onPause() {
...
         super.onResume();
         MainApp.bus().register(this);
-        sRefreshLoop = new Runnable() {
-            @Override
-            public void run() {
-                scheduleUpdateGUI("refreshLoop");
-                sLoopHandler.postDelayed(sRefreshLoop, 60 * 1000L);
-            }
-        };
-        sLoopHandler.postDelayed(sRefreshLoop, 60 * 1000L);
```
MilosKozak: why? this is here to update time

jotomo: You're right, that's stupid, I totally misread that.

PR #854, opened 2018-04-08
Correct lastDataTime handling

TebbeUbben:


AdrianLxM:
Thanks! I hope this will fix the "unreachable alert" issue with the insight :)

PR #857, opened 2018-04-08
Hide TDD buttons

TebbeUbben:
This hides the TDD button in actions if the pump doesn't support them and hides the reload button if the pump doesn't need this.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/Actions/ActionsFragment.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/Actions/ActionsFragment.java
@@ -79,8 +79,6 @@ public class ActionsFragment extends SubscriberFragment implements View.OnClickL
             tddStats = view.findViewById(R.id.actions_tddstats);
             history = view.findViewById(R.id.actions_historybrowser);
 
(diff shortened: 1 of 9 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/Actions/ActionsFragment.java
```diff
@@ -79,6 +79,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
             tddStats = view.findViewById(R.id.actions_tddstats);
             history = view.findViewById(R.id.actions_historybrowser);
 
+            if (!ConfigBuilderPlugin.getActivePump().getPumpDescription().supportsTDDs) tddStats.setVisibility(View.GONE);
```
AdrianLxM: Could you please move this to `updateGui()` with an if/else?
I guess that when changing the pump (to MDI for pairing a new Combo e.g.) it only would reload by coincidence (if it is far enough away from the config builder that the view gets destroyed).

TebbeUbben: Done.

PR #859, opened 2018-04-09
Watchsmb

AdrianLxM:
This seems to have been targeted to `watchsmb` after the merge and not to `dev`.
We might have forgotten that.


AdrianLxM:
Thanks @andrew-warrington

PR #865, opened 2018-04-10
Fix for black number and labels on black background (aaps wear)

swissalpine:
Something strange happened.
image
This could/would fix it ...


AdrianLxM:
Thank you very much!

PR #868, opened 2018-04-11
WIP: Carbs generator and treatment dialog updates

jotomo:
Overhauls the carbs dialog to add a duration field.
If set, carbs are split up and are generated every 15m over the specified period. This shall then trigger SMBs to deal with these carbs. The idea is to simulate how high fat/carb meals are absorbed by the body and give the loop hints, so it can dynamically bolus for it. The scenario is the same as using an extended/dual bolus.
This is an initial version which will see fixes and adjustments but I would be interested in feedback about the idea. @ceben80 already tried this with a pizza with good results.


jotomo:
screenshot_1523445862
screenshot_1523445906
screenshot_1523445913

MilosKozak:
👍 great approach
I'happy you are faster than me. I only started with food database to be able enter fat and proteins.

i'd have an idea but without looking at the code how complicated it would be
what about to implement `ExtendedCarbs`. We know how to do it from extened bolus. I'd use the same principle cutting to parts between BGs.
The andvantage would be only one record.
If i'm looking to the future, it would probably possible to shift expected start of ExtendedCarbs to time when we detect it from deviations or after common carbs absorbed.

Your thoughts?

AdrianLxM:
I really like the idea! For Low-Carb but also for scheduling carbs.

What we should check is, if it works with the pumps that store the carbs like the Dana*R and especially the RS as there the history is actively read.

This PR also includes the "carbs" field preset with the "carbsReq" from last determine basal run. I wonder if we should restrict that to a certain timeframe (e.g. "last run not older than 11 minutes"?).

@jotomo could you please mark it WIP until "you have found your clothes again"? ;)

jotomo:
@MilosKozak Current code is trivial, basically just a UI to small loop creating carb records to see if the idea plays out. `ExtendedCarbs` sound great and is easier to delete if needed. I like the idea of shifting the start based on deviations!

jotomo:
@AdrianLxM Already did. Wanted to see if someone opens the help dialog (and also just to have some silly fun) :-)

PoweRGbg:
Great idea and work! Can't wait to test it.

AdrianLxM:
Rounding Issues: When entering 50g of carbs over 8 hours it will enter 8 x 4 x 2g (=64g) of carbs. That's 28% more.

My idea would be to generate the rounded value and then for the next 15min event, calculate the rest and divide it through the remaining time and round again.

gregorybel:
Great idea!

AdrianLxM:
On b1f66d0 it does not creat too many carbs anymore, but now spreads it out to wide. 40g of carbs over 7 hours will lead to over 9 hours of 1g carb entries

AdrianLxM:
My "problems" are now resolved with 10b9171

jotomo:
I think I'm done tweaking and finding issues. Please review and merge if there are no more issues.

jotomo:
Code is broken for DanaR* pumps. Since they support storing carbs, the generated carbs are queued like boluses and thus most of them are discarded by the queue ....

Heiner1:
So DanaR* users should look to Milos's ideas about extended carbs or similar.

triplem:
I do like the notes field on all of those dialogs. We should move the "Show notes in treatment dialogs" switch beneath the buttons config, I guess. 
Furthermore, on the Carbs dialog, I do like the "+"-Buttons beneath the Carbs. But I do think, that those buttons could be mis-interpreted to be useable in all above input fields. We could add a separator (an empty line) in between the Duration and the Carbs. What do you think?

triplem:
Notes for Insuline is stored as well, which is the only dialog I did not touch concerning this one. Perfect.

jotomo:
* Ah, you mean to move the pref **into** the buttons menu. Yes, that's clearer, I'll do that.
* I added a separator line between the TTs and the carb fields to make it clearer they're separate: TTs always start now, only the carbs use the time/duration information. I thusly grouped time, duration and carbs together. Adding another separator makes that bit less clear. Though I see your point that this the +/- buttons might not be as intuitive for someone who hasn't seen previous versions of the dialog. 
* w.r.t. to DanaR* support: I'm tempted to just add carbs to the DB without sending it to the pump at this point. Not really sure how to proceed otherwise, as I can't test if the pump will store carbs in the future (don't have a DanaR) and I don't want to invest work into changing the queue to support multiple carb-only treatments since I consider this an experiment and the jury on real-world usefulness is still out (feature is only shown/enabled in engineering mode only).

MilosKozak:
I hope this not only experiment :) I wanted to fix a bug in InsulinDialog but it's conflicting too much with this code
fix this please 
```
                                .low(Profile.toMgdl(finalTT, currentProfile.getUnits()))
                                .high(Profile.toMgdl(finalTT, currentProfile.getUnits()));
```
storing carbs on DanaR is the only issue?

MilosKozak:
there is more updates needed. i'll create pr

AdrianLxM:
@MilosKozak before you start something, I'm currently working on the following: https://github.com/MilosKozak/AndroidAPS/commit/2561c3f8b6ae4bb09a1b3fc6d89b47562a1ef53b

MilosKozak:
ok, there are another bugs i'm working on

MilosKozak:
let's continue here https://github.com/MilosKozak/AndroidAPS/tree/carbs-gen-pr
hard to work on @jotomo 's repo

jotomo:
> I hope this not only experiment :)

Not any longer apparently, people seem to like it :-)

AdrianLxM:
Continuing here: https://github.com/MilosKozak/AndroidAPS/pull/892

Issue #841, opened 2018-04-05
add note to "AndroidAPS started" what phone was it


MilosKozak:

LongboatAline:
Good idea for all of us who are testing different setups. You'll use the ID that's also visible with `adb devices`, I assume? (I'm usually using the same kind of phone, so if they're all called "Moto G", they'd still be hard to tell apart)

PoweRGbg:
Should it be another note or just a string added to the AndroidAPS started ?

MilosKozak:
thanks @PoweRGbg

PR #869, opened 2018-04-11
Phone model note added to Ns #841

PoweRGbg:
right after the AndroidAPS Started note


app/src/main/java/info/nightscout/utils/NSUpload.java
```diff
@@ -476,6 +476,32 @@ public static void uploadAppStart() {
...
+        if (SP.getBoolean(R.string.key_ns_logappstartedevent, true)) {
+            Context context = MainApp.instance().getApplicationContext();
+            Bundle bundle = new Bundle();
+            bundle.putString("action", "dbAdd");
+            bundle.putString("collection", "treatments");
+            JSONObject data = new JSONObject();
+            try {
+                data.put("eventType", "Note");
+                data.put("created_at", DateUtil.toISOString(new Date()));
+                data.put("notes", "Started on "+ Build.MANUFACTURER + " "+ Build.MODEL);
```
MilosKozak: this would create another NS entry. only add this part `+ Build.MANUFACTURER + " "+ Build.MODEL` to previous event

PR #872, opened 2018-04-11
Notes wizard

triplem:
This PR adds a new Note field to the Wizard as well as the Carbs dialog. These fields are added to the DetailedBolusInfo, which is then transferred to NS. There is no additional data stored in AAPS.

bildschirmfoto von 2018-04-11 20-11-30

bildschirmfoto von 2018-04-11 20-11-43


triplem:
Right now, there is no Checkbox in the preferences to enable/disable those fields.

PR #873, opened 2018-04-11
Objective

### LINKED ISSUE #842: You can finish Objective 7 and 8, without full time elapsing
So when I entered objective 7 (and 8) I could proceed to finish without 20-something days to pass before (I am been using AAPS for time longer than that time required). I think problem is that objective doesn't look when you started it, but on full time you have been using AAPS. Same with 8.

Version: 1.58 Combo branch (I use my own fork, because I am working on adding translation, so build number won't be of much help, but if it does here it is - ad4795c 2018.03.18)

triplem:
Looks like the boolean condition was incorrect, but the problem never occured because of requirementsMet.done condition.


AdrianLxM:
Before merging: I do see a problem here. Will write in the channel.

triplem:
Cleaned and closed.

PR #876, opened 2018-04-12
upload temptargets to NS properly

MilosKozak:


jotomo:
@MilosKozak TTs via Careportal, watch, insulin- and carbs-dialog: all working as expected, uploading to NS :+1:

PR #884, opened 2018-04-16
More generous values for colored iage warnings (careportal)

swissalpine:
Normaly Insulin should be stable at least 4-5 days. So I think a warning after two days and a red alarm after three is two early. I changed the default values to 72 hours for the yellow and 86 hours for the red warning.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/Careportal/CareportalFragment.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/Careportal/CareportalFragment.java
@@ -216,8 +216,8 @@ public class CareportalFragment extends SubscriberFragment implements View.OnCli
                         CareportalEvent careportalEvent;
                         NSSettingsStatus nsSettings = new NSSettingsStatus().getInstance();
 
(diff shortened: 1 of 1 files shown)


MilosKozak:
these values are alligned to NS defaults

swissalpine:
Hm. What to do now?
The common sense in diabetics postulates other values ... A iage warning after 2 days makes no sense in my opinion.

AdrianLxM:
Yes, it seems that IAGE was just copied defaults from CAGE.

Ideas:
* Don't hard-code that values but make them adjustable settings?
* Do a PR at Nightscout to fix it there as well?

MilosKozak:
send PR to NS as well
@AdrianLxM it's not hardcoded, it's only default

swissalpine:
I couldn't calculate: I meant 96 hours ... (PR is updated)
I proposed this change related to the wiki text on http://www.nightscout.info/wiki/welcome/website-features/0-9-features also via the web formular on the Nightscout site: http://www.nightscout.info/contact
Is there another place for both: changing and proposing?

PR #891, opened 2018-04-18
Dana Queue for Carbs Generator

AdrianLxM:


AdrianLxM:
It works well to send the Carbs to the Dana*R pump.

My pump has them for the date "1/1/70 1:00". But other carbs usually get also logged that way so no change there.

PR #895, opened 2018-04-18
Cleanup mmol/mgdl TT handling in Dialogs

triplem:
Centralize the determination of mmol/mgdl default values for TempTargets


MilosKozak:
the carbsdialog part will be unmergeable with https://github.com/MilosKozak/AndroidAPS/tree/carbs-gen-pr

Issue #817, opened 2018-03-31
Meal Issues


AdrianLxM:
* (1) Treatments-Plugin counts carbs and "lastCarbTime" depending on DIA. If we have a "max absorption time", use that instead.

 * (2)`result.lastCarbTime` seems to be wrong if two meals fall in the same time. The oldest is recognized, not the newest.

 * (3) Use "max absorption time" to cutoff COB in `AutosensData#removeOldCarbs`

 * (4) Introduce the setting "max absorption time" also for the OREF sinsitivity plugin in order to have it in (1)&(2)

 * (5) How is UAM really handled after an announced meal. Wasn't there an exclucion of that timeframe?

 * (6) What to do in the OREF sensitivty plugin, when "max absorption time" and "min-impact" contradict each other?

 * (7) determine-basal gets the "min_5m_carbimpact" as parameter. So far it might get the default value that has nothing to do with the user-set "max absorption time" or even worse: Something that was put in the setting long before that is no longer visible in the settings while the AAPS-Sensitivty plugin is activated. This only affects "AMA". SMB does no longer use this parameter. We should check if any other parameter gets affected.

My opinion regarding (6): ignore that case. The "max absorption time" will be there to do the cutoffs. "5 min min impact" is the main setting here.

It would be nice if we could look at this in group. :)

AdrianLxM:
* (8) Multiple treatments decay too fast
In `info/nightscout/androidaps/plugins/IobCobCalculator/IobCobThread.java`, the minCarbImpact of all treatments gets summed up.

```
                        // calculate sum of min carb impact from all active treatments
                        double totalMinCarbsImpact = 0d;
                        for (int ii = 0; ii < autosensData.activeCarbsList.size(); ++ii) {
                            AutosensData.CarbsInPast c = autosensData.activeCarbsList.get(ii);
                            totalMinCarbsImpact += c.min5minCarbImpact;
                        }

```

This means if you enter a meal in 2 or more portions, the minCarbImpact will rise drastically.

PR #898, opened 2018-04-19
Please review: Cob rework

AdrianLxM:
* Don't cutoff at 4 hours but at max time for AAPS sensitivity
* For Oref sensitivity: new setting for cutoff
* Use the same `min5minCarbImpact` in all places. If COB is calculated dynamically, use that one. Use value defined in Constants/SMBDefaults.
* Don't sum up the carb impact for OREF sensitivity. Two meal entries directly after each other should behave like one larger meal.
* Sum up min carb impact for AAPS sensitivity, as the min carb impact for each meal depends on the carb size. The overall min impact should be the same as one larger meal.
* Use the absorption time not DIA for meals in the treatments plugin

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/IobCobCalculator/AutosensData.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/IobCobCalculator/AutosensData.java
@@ -65,7 +65,7 @@ public class AutosensData {
     public double autosensRatio = 1d;
     public double slopeFromMaxDeviation = 0;
     public double slopeFromMinDeviation = 999;
-    public double usedMinCarbsImpact = -0;
+    public double usedMinCarbsImpact = 0d;
 
 
     @Override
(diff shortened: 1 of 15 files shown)


AdrianLxM:
This should solve 1, 2, 3, 4, 7 and 8 of:
https://github.com/MilosKozak/AndroidAPS/issues/817
6 we could ignore IMHO, 5 is not related and should be done separately

Not tested yet.

app/src/main/java/info/nightscout/androidaps/plugins/IobCobCalculator/AutosensData.java
```diff
@@ -34,14 +36,14 @@ public CarbsInPast(Treatment t) {
...
-                double maxAbsorptionHours = SP.getDouble(R.string.key_absorption_maxtime, 4d);
+                double maxAbsorptionHours = SP.getDouble(R.string.key_absorption_maxtime, Constants.DEFAULT_MAX_ABSORPTION_TIME);
                 Profile profile = MainApp.getConfigBuilder().getProfile(t.date);
                 double sens = Profile.toMgdl(profile.getIsf(t.date), profile.getUnits());
                 double ic = profile.getIc(t.date);
                 min5minCarbImpact = t.carbs / (maxAbsorptionHours * 60 / 5) * sens / ic;
                 log.debug("Min 5m carbs impact for " + carbs + "g @" + new Date(t.date).toLocaleString() + " for " + maxAbsorptionHours + "h calculated to " + min5minCarbImpact + " ISF: " + sens + " IC: " + ic);
             } else {
-                min5minCarbImpact = SP.getDouble("openapsama_min_5m_carbimpact", 3.0);
+                min5minCarbImpact = SP.getDouble("openapsama_min_5m_carbimpact", SMBDefaults.min_5m_carbimpact);
```
MilosKozak: in xml there is default still 3.0

AdrianLxM: Thanks @MilosKozak ! Now they are the same :)

Issue #901, opened 2018-04-20
[dev] Disabling Auto Backfilling does not disable Loop on new SGV


AdrianLxM:
... what annuls the function to block values on too high noise in xDrip in order to not trigger SMB.

jotomo:
Sorry, the PR doesn't fix that: whether `SourceXdrip` claims `advancedFiltering` is supported depends on the last _retrieved_ BG value (see `setSource` method), which isn't necessarily the one being looped with.
The more I think about it (including processing noise, filtering in xDrip etc), the more I think that `advancedFiltering` isn't sufficent, but that we need new data in `BgReading` stating whether the value is suited to SMB with.
I think this needs to be thought through more extensively; oref 0.6.1 raises the target on noise, rather than disable SMB. We might also want to disable the loop entirely if noise gets excessive. And if we add a plugin to do source-plugin independant noise filtering, we need to inspect more than just one value (but might still add the computation result as something like `BgReading.allowSMB`).

jotomo:
@AdrianLxM should the title read **enabling** backfilling? The _LoopPlugin_ now only loops for recent readings and readings newer than the last reading the loop ran for (see LoopPlugin:145).
The issue of the loop being invoked with high-noise data would occur if the most recent value had high noise (=not G5 native) and then G5 native backfilled readings came in setting _advancedFiltering_ to _true_ (due to the user switching sources in xDrip to G5 native). Even then, the synchronous nature of the event bus (backfilled readings processed last) would prevent looping with a high noise value. Since other sources other than G5 don't (at this time) support backfilling, a user switching from G5 native to another source shouldn't be an issue. Does that sound reasonable?

AdrianLxM:
There are a couple of sources that allow backfilling:
* Libre->xDrip
* Probably Glimp
* G5-xDrip
* Nightscout.

Nightscout is tricky as it can be anything and a second path for data.

Can you please check that data coming from Nightscout does not trigger a loop run if it is not enabled as source. One of the issues was: xDrip blocked the local broadcast to AAPS due to high noise. It still uploaded to Nightscout and therefore AAPS continued looping on that noisy data.

AdrianLxM:
Will be solved on the xDrip side. Thank you Jon!

Issue #671, opened 2018-02-06
Multiple loop invocations using xDrip as source and uploader to Nightscout, when NSClient is enabled


jotomo:
When xDrip is used as a source for SGVs (receiving local broadcasts) and xDrip is also configured to upload SGVs to Nightscout, while AAPS uses NSClient the following happens:
AAPS receives the new SGV locally from xDrip, fires `EventNewBg`, which triggers IOB & COB calculations, which in turn trigger a loop run. xDrip uploads the SGV and NSClient receives that SGV shortly after. The existing SGV (received from xDrip) is updated with the record received via NSClient and `EventNewBg` is fired again, triggering the above again, 5-10s after the first run. 

Should the loop be invoked by SGV data coming in from a different source than what is configured as "BG source"?

jotomo:
Same solution as with #694 via https://github.com/MilosKozak/AndroidAPS/commit/e9313c5b2849a646ca703b08bce32abae4f8bb8d.

jotomo:
Not sure yet how to do this exactly (updating UI etc without triggering the loop), other than embedding the source in EventNewBg so that LoopPlugin can check if the BG comes from the active BG source. Should also check the retrieved BG is not older than 5m or so, so it won't trigger a loop run when retrieving back-filled data.

MilosKozak:
I was thinking about recording bg time from last EventNewBg . what do you think?

jotomo:
... local state that is lost on restart (negligible), but other plugins also need to differentiate between backfilled BGs and new BGs.
I like the straightforwardness of an Event providing the data needed to make an informed decision, hence my counter-proposal :-) : 
https://github.com/MilosKozak/AndroidAPS/compare/dev...jotomo:new-bg-processing?expand=1

MilosKozak:
why not. make PR

jotomo:
@MilosKozak I should have just listened to you ;-) #906

PR #906, opened 2018-04-21
Don't loop with old BG value, don't loop twice for the same value.

jotomo:
Loop is potentially triggered twice when BG reading sent to NS comes back.
This should also deal with backfilled data coming in, since any previous
reading will be older than 9m, for which DatabaseHelper.actualBg()
returns null.

The previous approach to solve multiple invocations for the same value
added a isNew flag to EventNewBG, but since
DatabaseHelper.scheduleBgChange() drops excessive updates the event
carrying that flag is not guarantued to be delivered, resulting in
missed loop invocations. The approach taken now lets the receiver
of the event fully decide how to deal with it.

Should fix, or at least improve, #901, #671.


jotomo:
WIP, I need to test this.

jotomo:
Tested with Nightscout now.

MilosKozak:
ready?

jotomo:
Yes

PR #907, opened 2018-04-21
Compatiblity for SR 1.5.8 and add 0%-TBR to history when pump has been paused or stopped

TebbeUbben:


TebbeUbben:
What happens in case there are multiple TBRs overlapping each other?

E.g.:
10:00: 90% for 60 min
10:10 Pump paused for 10 min (so 0% for 10 min)
10:20 90%-TBR continues running

MilosKozak:
TBR will not continue. you can add an example to tests

PR #908, opened 2018-04-21
Fix delay in COB calculation.

jotomo:
Line 110: The time of the newly received SGV is rounded up to a full minute, e.g. from 23:43:20 to 23:44. This is compared against `System.currentTimeMillis`, which will be around 23:43:20. Thus, bgTime evalutes to being greater than 'now' and the iteration is skipped for this latest SGV, not updating COB.
This change rounds 'now' up to a full minute as well, so that the comparison works.
There is still a (small) chance of the issue occurring when a minute is crossed between the time an SGV reading is captured to when COB calculation starts.


MilosKozak:
good catch 👍

PR #912, opened 2018-04-22
Add IobCobCalculatorPlugin.getCobInfo().

jotomo:
* COB on overview is now always up-to-date and includes carbs not yet processed by IobCob calculation. 
* Same for the wizards, so that two entered meals between calculations have proper input.
* All scheduled/future carbs are displayed in brackets (if any) with COB on overview.

Roughly tested. Calculation of `displayCob` needs a check to ensure carbs aren't added which are already processed.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/data/QuickWizardEntry.java
+++ b/app/src/main/java/info/nightscout/androidaps/data/QuickWizardEntry.java
@@ -11,7 +11,7 @@ import info.nightscout.androidaps.R;
 import info.nightscout.androidaps.db.BgReading;
 import info.nightscout.androidaps.db.TempTarget;
 import info.nightscout.androidaps.interfaces.TreatmentsInterface;
-import info.nightscout.androidaps.plugins.IobCobCalculator.AutosensData;
+import info.nightscout.androidaps.plugins.IobCobCalculator.CobInfo;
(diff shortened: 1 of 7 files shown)


jotomo:
If OpenAPS plugin is invoked manually, it will NOT incorporate carbs entered after the last BG, which might be confusing. Not sure if something should be changed there.

jotomo:
@MilosKozak What do you think of this?

jotomo:
Not reday.

jotomo:
@AdrianLxM Please review.

AdrianLxM:
When a new treatment is entered, a new loop-run will be triggered as well. Will this one see the COB already?

jotomo:
Good one. No ... same is with a user triggered loop run. Would passing in updated COB (with everything else Autosens not being up-to-date) work?

AdrianLxM:
@jotomo 
> Good one. No ... same is with a user triggered loop run. Would passing in updated COB (with everything else Autosens not being up-to-date) work?

There is another field `Mealdata->carbs` that determine-basal uses apart from COB. It is the overall carbs that were considered to calculate the COB. It should work if the carbs would get added there too.

jotomo:
Do you still remember from the absorption work how that area of the code works? I'm unfamiliar with it and if you leave it to me I'm likely gonna break something (and then some) in the process.

PR #914, opened 2018-04-22
Overview: hide ext. bolus line if none running; spacings, alignments.

jotomo:


app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
```diff
@@ -1174,7 +1174,7 @@ public void onClick(View v) {
                 extendedBolusView.setText(extendedBolusText);
             }
             if (extendedBolusText.equals(""))
-                extendedBolusView.setVisibility(View.INVISIBLE);
+                extendedBolusView.setVisibility(View.GONE);
```
MilosKozak: GONE breaks formating of ":". 
or is it fixed below?

jotomo: Below is only spacing and alignments.
Without EB running:
image
With EB running:
image

jotomo: What ":" do you mean?

PR #924, opened 2018-04-24
Is profileswitchdirect necessary (dialog overview fragment)?

swissalpine:
I'm missing the possibility to enter event time in the profile switch dialog started via the overview fragment. So I propose to eliminate the profile switch direct alternative because I couldn't see the benefit to save up the possibility to enter an event time (only one line plus).


swissalpine:
Oh. Doesn't work yet. Closed.

MilosKozak:
the idea is to have on overview only the most used dialogues with as less controls as possible. only to cover daily routine

swissalpine:
Yes. I know. But this is not my daily, but regularly practiced routine. With TT dialog on overwiew tab I could enter a start time ... And I really missed this feature, it's (for me) more important as entering notes - but yes, it's possible to start this from the careportal tab.
What I don't understand is why my commits ignore the percentage if I schedule it. On the careportal it works as it should.

Issue #681, opened 2018-02-10
SMBs not delivered on DanaR


AdrianLxM:
Problem: SMB of 0.4 units gets issued, delivered only 0.0.
**This happened multiple times, afair over an hour - only a phone restart fixed it**

Here the combination of TBR and SMB gets issued
```
21:32:14.908 [Thread-3213] DEBUG [info.nightscout.androidaps.plugins.ConfigBuilder.ConfigBuilderPlugin:382]: applyAPSRequest: Rate: 0.00 U/h (0.00%) 
Duration: 30.00 min
SMB: 0.40 U
Reason: COB: 63.77024083689827, Dev: 46, BGI: -4, ISF: 47, CR: 8.36, Target: 100, minPredBG 226, minGuardBG 166, IOBpredBG 79, COBpredBG 370;  insulinReq 2.71; maxBolus 0.4; setting 30m low temp of 0U/h. Microbolusing 0.4U. 
```

...

And here the result in the QUEUE is "OK" but only 0.0 of 0.4 was given.
```
21:32:19.906 [Thread-3214] DEBUG [info.nightscout.androidaps.queue.commands.CommandTempBasalAbsolute:35]: setTempBasalAbsolute rate: 0.0 durationInMinutes: 30 success: true enacted: true
21:32:20.011 [Thread-3214] DEBUG [info.nightscout.androidaps.queue.QueueThread:120]: QUEUE: performing SMBBOLUS 0.4U
21:32:20.014 [Thread-3214] DEBUG [info.nightscout.androidaps.plugins.PumpDanaR.DanaRPlugin:145]: deliverTreatment: OK. Asked: 0.4 Delivered: 0.0
```

As SMB or replace basal insulin and TBRs are scheduled as if SMB would work, this is a rather bad situation. (ALARM?)

Full log:
[AndroidAPS._2018-02-09_22-07-46_.0.zip](https://github.com/MilosKozak/AndroidAPS/files/1713883/AndroidAPS._2018-02-09_22-07-46_.0.zip)

AdrianLxM:
Disregarding the problem **WHY** delivering a treatment failed (@MilosKozak it would be nice if you could find a reason for that), there are a few more questions:

* Is it good to add to treatment even if the `result.success` is `false`
* With SMB we can no longer disregard if one of the two actions that are issued simultaneously failes and one has success. Only if both fail, we are in the failsafe state "plain old insulin pump delivering basal insulin".

[...]

PR #925, opened 2018-04-24
Safety loop

AdrianLxM:
To make it more easier to grasp what my intention in https://github.com/MilosKozak/AndroidAPS/issues/681 was.

What would you think of an approach like this?

I'm not 100% sure if starting the safety-invoke on a new Thread like this will work. It should be on another thread as the callback happens on the CommandQueue-Thread.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/interfaces/APSInterface.java
+++ b/app/src/main/java/info/nightscout/androidaps/interfaces/APSInterface.java
@@ -11,5 +11,5 @@ public interface APSInterface {
     public APSResult getLastAPSResult();
     public Date getLastAPSRun();
 
-    public void invoke(String initiator, boolean safety);
+    public void invoke(String initiator, boolean tempBasalFallback);
 }
--- a/app/src/main/java/info/nightscout/androidaps/interfaces/APSInterface.java
+++ b/app/src/main/java/info/nightscout/androidaps/interfaces/APSInterface.java
@@ -11,5 +11,5 @@ public interface APSInterface {
     public APSResult getLastAPSResult();
     public Date getLastAPSRun();
 
-    public void invoke(String initiator);
+    public void invoke(String initiator, boolean safety);
 }
--- a/app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
@@ -417,6 +417,9 @@ public class ConfigBuilderPlugin extends PluginBase {
      */
     public void applyTBRRequest(APSResult request, Profile profile, Callback callback) {
         if (!request.tempBasalRequested) {
+            if (callback != null) {
+                callback.result(new PumpEnactResult().enacted(false).success(true).comment(MainApp.gs(R.string.nochangerequested))).run();
+            }
             return;
         }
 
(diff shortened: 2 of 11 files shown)


MilosKozak:
what is the `safety` for?

AdrianLxM:
It will disable SMB in the run. This way on the bug where no SMB is delivered it will fall back to Oref0 and not give the 0-TBR where it expects there are SMBs

AdrianLxM:
On my test-run:
eventualBG: 122mg/dl
target: 90mg/

At first it set a 30% TBR.
Then it tried to do a SMB of 0.2U - failed
-> safety run with SMB deactivated
Then it set a TBR of 260%.

So it actively prevents low-temping when actually more insulin should be given (but bolusing fails).

AdrianLxM:
In addition: Starting the thread like this works! :)

jotomo:
Thoughts, no solutions:
* 'safety' is not a good name to convey the intent. _forceTbrOnly_, _preventSmb_, _tempBasalFallback_?
* Might it be clearer to pull the `tempBasalRequested` and `bolusRequested` checks into LoopPlugin? (Someone called me did a lazy job there)
* Thread.sleep() and concurrent invocations, from a new BG (Combo is slow if e.g. TBR/SMB has to wait for a pump basal profile update if a profile switch ended), manual invocations from the user ... some synchronization needed?

AdrianLxM:
@jotomo 
1) I renamed "safety" to "tempBasalFallback". Good idea

2) Could you elaborate a bit please.

3) The `SystemClock.sleep(1000);` only adds one more second. So I wouldn't focus on that. But you are right, there might be "parallel" loop invocations. Not only now but already before with a manual loop trigger, a treatment change event, ... just with a lesser likelihood.
Do you think it suffices to synchronize `LoopPlugin#invoke()`?

jotomo:
2) & 3) are refactorings which also touch this PR, but were there before.

2) The `ConfigBuilder.applyTBR/SMBRequest` methods return immediately, without invoking the callback, when `request.tempBasalRequested` or `request.bolusRequested` are false. This was ok-ish before, but now that it matters more whether one of those is actually executed, it might make the code clearer to pull those checks from those methods and have `LoopPlugin.invoke` do those checks and only invoke the respective method if we know we have work for them (like is done in ~335 to set the `waiting` state for the UI).
3) I think synchronization will be sufficient (A `synchronized` method can invoke itself since it holds the lock, right?). Adding some _cancel an earlier request if we have a new one_ logic would likely be hard to reason about and prone to errors. With synchronization the loop will run multiple times in short succession, but that shouldn't happen too often and does no damage.

AdrianLxM:
Ok, I synchronized the loop invocation. Yes, a synchronized method can call another one (on the same object) and therefore also itself recursively, as both synchronize to `this` and therefore hold the lock. In this case it would even be the callback thread, that does that.

On the other topic: I think it is ok, if `ConfigBuilder.applyTBR/SMBRequest` return without callback if there is no request. But we should check if there is the possibility of a SMBRequest without a TBR request.

AdrianLxM:
@jotomo so this should do it: https://github.com/MilosKozak/AndroidAPS/pull/925/commits/e79ef7f0186c759d99b5eedc2931c4da7f0e3feb ?

jotomo:
Yes.

AdrianLxM:
No invocation on treatment works fine.
Jotomo already used this for a while.

PR #929, opened 2018-04-25
Proposal: Change iob and cob color for a better uniform look.

osodebailar:
Pleas review and if you like it merge it.
screenshot_20180424-194323

Insulin , iob etc. always in different blue colors,
Cob in orange (ocher) like color


AdrianLxM:
Thanks @osodebailar! merging...
(If there are any objections later, easy to undo as it does not change any executed code)

PR #931, opened 2018-04-26
align failoverToMinAbsorbtionRate to cobtime

AdrianLxM:
The time on the chart and the time of the autosensedata did not align. Autosensedata usually is earlier as the one before gets picked.

before:
screenshot_20180426-030307

after:
screenshot_20180426-031238


AdrianLxM:
The transparency of the COB background was "B0..." instead of "80..." like the comment said (50%). I also changed that.
screenshot_20180426-034246

Issue #932, opened 2018-04-26
IOB-Array-Issues


AdrianLxM:
We should check `IobCobCalculatorPlugin#calculateIobArrayForSMB()`:

 * I've noticed, that ZT in determine basal always looks quite short.

 * Also it still uses `int len = (4 * 60) / 5;` no matter what the actual DIA is.

 * `calculateFromTreatmentsAndTemps(long time, Profile profile)` seems to not  truncate the currently running themp. So for `basalIob.iobWithZeroTemp` it might calculate the currently running temp into the future plus the 0-temp. I believe this value should express "what if determine-basal would set a 0-temp right now and stop the current one".

AdrianLxM:
The shortness of ZT is ok - they get shortened by determine basal as intended

The Lenght might not be as defined with the DIA, but determine-basal will only consider the first 48 anyway - that's two hours in 5 min intervals

Only `calculateFromTreatmentsAndTemps(long time, Profile profile)` remain.

AdrianLxM:
Proposed fix: https://github.com/MilosKozak/AndroidAPS/pull/934

PR #934, opened 2018-04-26
For review: Future iob array

AdrianLxM:
Treat future IOB as if the current TBR was cancelled


AdrianLxM:
If it works as expected, it should fix: https://github.com/MilosKozak/AndroidAPS/issues/932

MilosKozak:
good catch !

PR #947, opened 2018-05-01
Don't create empty future carbs records. Fixes #940.

### LINKED ISSUE #940: [RC1] Extended carbs creates empty Carb correction entries in Nightscout
We entered extended carbs for 10g over a duration of 4h. While creating the correct amount of 1g carb corrections, AAPS also creates unnecessary empty treatments.
ns1
ns2

jotomo:
When very few carbs are stretched out (e.g. 10g over 4h), there are
gaps, so that 1g is generated only every 30m. Don't create empty
records in between.


jotomo:
screenshot_1525170516
@AdrianLxM Unintended (I presume) but nice side effect: towards the end, the gap between carbs increases as there are less and less carbs to distribute, which should fit nicely with how carbs are absorbed, no?

philipgo:
@jotomo A small suggestion: The carbs are labeled as Carb Correction, could you change that to Meal Bolus?

jotomo:
@philipgo Just extended ones? Or all?
@MilosKozak there was a reason why AAPS creates them with type `Carb Correction`. I don't remember why.

philipgo:
@jotomo It only caught my eye today when I saw lots of carb corrections in my reports. The extended ones are clearly never corrections since they are planned, so another label might be more suitable and raise less questions when looking at the reports with heaps of "carb corrections" all night. For everything else it does not seem too important.

jotomo:
@philipgo eCarbs will be created with type Meal Bolus via  #1001

philipgo:
@jotomo Great, thank you!

PR #952, opened 2018-05-01
Fix missing predictions in graph, color them.

### LINKED ISSUE #949: [RC1] Missing prediction lines on graph
Graph on overview sometimes only shows 1 or 2 predictions lines. Watch shows all three. Traced data to the point where the data series is added to graph and all data was present. So the guess is it's somewhere in the charting lib or our customizations.
Zooming out sometimes shows more lines.

jotomo:
image

Fixes #949


AdrianLxM:
screenshot_20180501-200532
Works great!

AdrianLxM:
... now we only need captions.

PR #956, opened 2018-05-01
rc2

islagroves:


AdrianLxM:
Hi, we will do this on master, when the release version of AAPS2.0 is ready.
Thanks

PR #962, opened 2018-05-02
refactor cob string

AdrianLxM:
Don't use a static method.
Correct reason, not always "WatchUpdaterService"


jotomo:
Tested, :+1:

PR #980, opened 2018-05-02
WIP: DE translation

triplem:


AdrianLxM:
Still WIP?

triplem:
Just used that translation editor. That not correct?

triplem:
formatted="false" is not necessary, if the positional format strings are used. I used this now in the german translation on some and added the formatted="false" on others as well, just to be on the safe side.

app/src/main/res/values-de/strings.xml
```diff
@@ -659,16 +659,16 @@
     <string name="btwatchdog_title">BT Watchdog</string>
     <string name="DexcomG5">DexcomG5 App (patched)</string>
     <string name="combo_pump_activity_label">Aktivität</string>
-    <string name="combo_tbr_remaining">%d%% (%d Min. verbleibend)</string>
-    <string name="combo_no_pump_connection">Keine Verbindung zur Pumpe seit %d Min.</string>
+    <string name="combo_tbr_remaining">%%1$d%% (%2$d Min. verbleibend)</string>
```
MilosKozak: this seems to be wrong

triplem:
well spotted

PR #982, opened 2018-05-02
 Combo: better error message when initialization fails

jotomo:
Bonus commit: revert change that dropped invalid boluses from being sent to watch. There we want them for display.


app/src/main/java/info/nightscout/androidaps/plugins/Wear/wearintegration/WatchUpdaterService.java
```diff
@@ -465,7 +465,7 @@ private void sendBasals() {
 
         List<Treatment> treatments = TreatmentsPlugin.getPlugin().getTreatmentsFromHistory();
         for (Treatment treatment : treatments) {
-            if (treatment.isValid && treatment.date > startTimeWindow) {
+            if (treatment.date > startTimeWindow) {
```
MilosKozak: strange. why it wants to remove it?

MilosKozak: oops. sorry

Issue #863, opened 2018-04-10
Virtual pump support


MilosKozak:
PUMP TYPE:
Bolus step:
Extended bolus step:
Extended bolus duration step:
Extended bolus max duration:

Temp basal style: (percent or absolute)
Temp basal step:
Temp basal duration step:
Temp basal max value:
Temp basal max duration:

Base basal min value:
Base basal step:

gempickfordwaugh:
PUMP TYPE: Animas Vibe https://www.animascorp.co.uk/insulin-pumps/vibe-insulin-pump
Bolus step: 0.05u
Extended bolus step: [not stated, assume as above]
Extended bolus duration step:  0.1 hr to 12 hrs, with 0.5 hr increments for 0.5 hr to 12 hrs
Extended bolus max duration: 12 hrs
Temp basal style: percent
Temp basal step: -90% to +200%, in 10% increments
Temp basal duration step: 0.0 hr to 24 hrs in 0.5 hr increments
Temp basal max value: 50u/hr (+200% of max basal rate 25u/hr)
Temp basal max duration: 24 hrs
Base basal min value: 0.025u/hr
Base basal step: 0.025u/hr steps

slon77:
PUMP TYPE: Accu-Check-Spirit-Combo
Bolus step:  0,5
Extended bolus step:  0,1 
Extended bolus duration step:  15 min – 12 h
Extended bolus max duration:  12 h
Temp basal style:  0 – 250 %
Temp basal step:  10 %
Temp basal max value:  500 %
Temp basal max duration:  24
Base basal min value:  0.01
Base basal step:  0,01 (до 1,00 U / h),
                           0,05 (до 10,0 U / h) и
                           0,1 (до 50,0 U / h)

gempickfordwaugh:
PUMP TYPE: Omnipod https://www.myomnipod.com/sites/default/files/inline-files/17845-5A%20Guide%2C%20Eros%20US%20User%20Guide%20Rev%20B.pdf
Bolus step: 0.05u
Extended bolus step: [not stated, assume as above]
Extended bolus duration step:  0.5hrs to 8 hrs, with 0.5 hr increments
Extended bolus max duration: 8 hrs
Temp basal style: percent
Temp basal step: 0u/hr to +95% in 5% increments
Temp basal duration step: 0.5hr increments
Temp basal max value: cannot exceed max basal rate 30u/hr
Temp basal max duration: 12hrs
[...]

PR #989, opened 2018-05-03
Virtual pump #863

andyrozman:
Implementation for Virtual Pump - Issue #863


andyrozman:
Sorry. I made my branch out of master, instead of dev. I will fix it, and then open another PR.

PR #992, opened 2018-05-03
WIP: Respect Meal times in autosense

triplem:


app/src/main/java/info/nightscout/androidaps/plugins/SensitivityAAPS/SensitivityAAPSPlugin.java
```diff
@@ -96,11 +98,24 @@ public AutosensResult detectSensitivity(long fromTime, long toTime) {
...
+            }
+
+            if (autosensData.time > toTime - T.hours(hoursForDetection).msecs()
+                    && !autosensData.nonCarbsDeviation)
                 deviationsArray.add(autosensData.nonEqualDeviation ? autosensData.deviation : 0d);
-            if (deviationsArray.size() > hoursForDetection * 60 / 5)
-                deviationsArray.remove(0);
 
+            // remove the deviation if the cob is only 45m behind current autosens time
+            if (deviationsArray.size() > hoursForDetection * 60 / 5
```
jotomo: `deviationsArray` can be empty (on start), causing an _IndexOutOfBoundExceptions_ when branch is entered (presumably due to the second condition being true)

AdrianLxM:
The issue this would fix has been addressed in the Autosens-rewrite.

PR #994, opened 2018-05-03
WIP: Startupwizard

MilosKozak:


AdrianLxM:
After pressing "finish" in the wizard, it is still o the backstack. -> It will open again when pressing "back".

PoweRGbg:
Also when esectinx Exit from the menu -> returns to the wizz

PR #997, opened 2018-05-04
Make deviation bars transparent

swissalpine:
image


jotomo:
Looks nice :+1: 
For me, the grey is still solid though. Setting an opacity of 200 makes it look almost like your screenshot (a bit too transparent though).
There are also black bars, setting an opacity of 114 there looks like this:
screenshot_20180504-112727

jotomo:
Although, grey are deviations which are ignored due to carb absorption, so at the time they're shown, COB & IOB are likely there and are more relevant, so having grey deviations not obfuscate IOB/COB too much is good.

jotomo:
Color definitions used for the screenshot:
```

swissalpine:
Thanks a lot for improving the commit. I integrated your proposals! 👍

jotomo:
@swissalpine Thanks!

Issue #939, opened 2018-04-29
[RC1] Bolus progress dialog hangs


jotomo:
Somewhat recently the code was removed to keep the display on while a bolus was delivered.
When the display turns off while a bolus is in running and is turned on again after the bolus has finished, the dialog hangs. Could it be that when the bolus command submits _EventDismissBolusprogressIfRunning_ that the dialog isn't dismissed, because it's not _running_ (not displayed) and therefor the dismissal never happens?
Maybe in the _BolusProgressDialog_ set a flag _dismissPending_  (in the handler) if not currently displayed? We need to make sure we're not accidentally dismissing the dialog for a follow up bolus (if that's a possibility, possibly when issuing a bolus from the watch after a bolus from the phone).

AdrianLxM:
The CommandQueue does not allow a new bolus if currently a bolus command is running.
The CommandQueue removes all bolus commands that are waiting in a queue when a new one gets entered.

If there are no race conditions with checking these conditions (running, remove prior) it should not be possible to have two boluses at once in the queue. (synchronization needed?)

jotomo:
Thanks, forgot about that.
I'll need to dig deeper, can't easily reproduce the behaviour right now.

jotomo:
```
java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
	at android.support.v4.app.FragmentManagerImpl.checkStateLoss(FragmentManager.java:2055) ~[na:0.0]
	at android.support.v4.app.FragmentManagerImpl.enqueueAction(FragmentManager.java:2078) ~[na:0.0]
	at android.support.v4.app.BackStackRecord.commitInternal(BackStackRecord.java:678) ~[na:0.0]
	at android.support.v4.app.BackStackRecord.commit(BackStackRecord.java:632) ~[na:0.0]
	at android.support.v4.app.DialogFragment.dismissInternal(DialogFragment.java:223) ~[na:0.0]
	at android.support.v4.app.DialogFragment.dismiss(DialogFragment.java:190) ~[na:0.0]
	at info.nightscout.androidaps.plugins.Overview.Dialogs.BolusProgressDialog.dismiss(BolusProgressDialog.java:86) ~[na:0.0]
	at info.nightscout.androidaps.plugins.Overview.Dialogs.BolusProgressDialog.lambda$null$2$BolusProgressDialog(BolusProgressDialog.java:151) ~[na:0.0]
[...]

PR #1000, opened 2018-05-04
IDEA: dismiss bolus progress dialog when Queue is empty

AdrianLxM:
As my phones do not show that behaviour described in https://github.com/MilosKozak/AndroidAPS/issues/939, I cannot test it.

Please discuss in issue #939


AdrianLxM:
Tested it on my old MotoG where I had this problem from time to time and I couldn't see it anymore.
As it only happened from time to time, this is not definitive.

PR #1007, opened 2018-05-04
eCarb also for small watches

swissalpine:
Unfortunatly I couldn't benefit from the new eCarb dialog on my Polar M600 (screen size 240x240):
screen
So I made changes: Reducing the font size of the labels and shorten the labels `duration [h]` and `start [min]`.
screen 1
Another possibility is to reduce the dimension of `+` and `-` buttons.


AdrianLxM:
As discussed in chat, there are other input designs that fit small and rectangular screens better.

Issue #1013, opened 2018-05-05
[rc3] cut off column -> "action/TDD/exponentially weighted TDD" in german translation


soko78:
In the tab "action/TDD" the column "exponentially weighted TDD (exponentiell gewichtete TDD)" is cut off on the right side.
![xdrip12](https://user-images.githubusercontent.com/34250545/39662886-5c866d20-5069-11e8-8550-e020a0def522.jpg)

AdrianLxM:
Fixed in https://github.com/MilosKozak/AndroidAPS/commit/260e259fb0f7e44dc5601a91aaf91c327409da0a

PR #1014, opened 2018-05-05
German Translation 'Tägl. Gesamtdosis' -> 'TDD'

swissalpine:
Fixing issue https://github.com/MilosKozak/AndroidAPS/issues/1013
[rc3] cut off column -> "action/TDD/exponentially weighted TDD" in german translation


AdrianLxM:
Thanks, sorry, just pushed exactly the same change :)

PR #1015, opened 2018-05-05
Fix finding BG in graph for treatments before the first displayed BG.

jotomo:
(cherry picked from commit 8bdafe3)


jotomo:
Had that thought as well, but didn't think it was worth the work; if there is no BG reading at all, the graph isn't really useful, is it?

Issue #653, opened 2018-02-03
Multi-line notifications are cut off


jotomo:
Notifications on the home screen have their last line cut off for longer (multi-line) messages

AdrianLxM:
Is this still an issue?

jotomo:
Yes, still an issue. Add
```
bus().post(new EventNewNotification(new Notification(999, gs(R.string.bolus_frequency_exceeded), Notification.NORMAL)));
```
to the end of `MainApp.onCreate` to reproduce on start:
![screenshot_1525433411](https://user-images.githubusercontent.com/1732305/39625797-723a94f4-4f9f-11e8-88cc-27c63beb9f6e.png)

AdrianLxM:
![screenshot from 2018-05-05 21-48-13](https://user-images.githubusercontent.com/9692866/39667046-13e1fe96-50ae-11e8-9e09-de43cedf62b6.png)
It seems to always just cut off the last line.

jotomo:
Fixed via #1037, thanks @AdrianLxM

PR #1017, opened 2018-05-05
Tests required: Notification height

AdrianLxM:
This is a possible fix for: https://github.com/MilosKozak/AndroidAPS/issues/653

As it cuts off the last line, it uses the "trick" of adding one line.
I did only test it in the simulator and not on a smaller phone. Please test and wait for tests before merging.

screenshot from 2018-05-05 21-55-23


jotomo:
Two real phones:
screenshot_20180505-222819
screenshot_20180505-222932

PR #1019, opened 2018-05-06
Obj.7 add info about using SMB/maxIOB

McHoffi:


AdrianLxM:
As discussed in chat:

1) It would be good to have this explanation in a separate String. This way we make sure to have this info in all translations as well (either translations or original, but not forgotten).

2) It will affect earlier Objectives as well when a manual bolus is given. As it is now factored in into the general IOB, it will block high TBRs - even without microboluses.

3) It is only affective, if "SMB" is selected as APS version.

McHoffi:
Would it maybe better to write also "a good start is maxIOB=average mealbolus + 3 x max daily basal"?

MilosKozak:
yes add it pls

McHoffi:
Done 👍

PR #1025, opened 2018-05-06
Allow showing 3 hour overview charts

MarkMpn:
X-axis labels on the overview chart can be misleading at the moment as only the hour part is shown, but the line can be on a half-hour. Some intervals can then seem to be 1 hour while the next one looks like 2 hours.

This change keeps the shorter `HH` format where possible and extends out to `HH:30` where it falls on a half hour.

Without this change the chart below is labelled `13 14 16 17 19 20 22`

screenshot_20180506-194855

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
@@ -314,8 +314,8 @@ public class OverviewFragment extends Fragment implements View.OnClickListener, 
             bgGraph.setOnLongClickListener(new View.OnLongClickListener() {
                 @Override
                 public boolean onLongClick(View v) {
-                    rangeToDisplay += 6;
-                    rangeToDisplay = rangeToDisplay > 24 ? 6 : rangeToDisplay;
+                    rangeToDisplay += 3;
+                    rangeToDisplay = rangeToDisplay > 24 ? 3 : rangeToDisplay;
                     SP.putInt(R.string.key_rangetodisplay, rangeToDisplay);
                     updateGUI("rangeChange");
                     return false;
--- a/app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
@@ -314,7 +314,12 @@ public class OverviewFragment extends Fragment implements View.OnClickListener, 
             bgGraph.setOnLongClickListener(new View.OnLongClickListener() {
                 @Override
                 public boolean onLongClick(View v) {
-                    rangeToDisplay += 3;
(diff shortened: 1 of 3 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
```diff
@@ -314,8 +314,8 @@ else if (dm.densityDpi <= 560)
             bgGraph.setOnLongClickListener(new View.OnLongClickListener() {
                 @Override
                 public boolean onLongClick(View v) {
-                    rangeToDisplay += 6;
-                    rangeToDisplay = rangeToDisplay > 24 ? 6 : rangeToDisplay;
+                    rangeToDisplay += 3;
+                    rangeToDisplay = rangeToDisplay > 24 ? 3 : rangeToDisplay;
```
jotomo: This doesn't change the default of 6h, but will increase the displayed timeframe by 3h for each long press, before wrapping around to 3h. This requires 7 long-presses - quiet a journey ...

jotomo:
I don't see the issues of markers not being evenly distribute on hour marks (or 2, 3 hours each if increasing timeframe). I don't find the display with half-hours especially easy to read. If there as an issue with the markers not on hour markers I think that should be fixed instead.

AdrianLxM:
What would you think of a multi-tap zoom like on the smartwatch?

jotomo:
Are 3h really useful?

Locally I have single-tap to increase range and long press to reset to 6h (multiple long or double-taps annoy me)
Can we catch up/down swipe events to increase/decrease the range and have long tap or double tap to reset?



On May 7, 2018 11:41:17 AM GMT+02:00, AdrianLxM <notifications@github.com> wrote:
>What would you think of a multi-tap zoom like on the smartwatch?
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/MilosKozak/AndroidAPS/pull/1025#issuecomment-387013875

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

AdrianLxM:
A single tap on the largest Element on the screen causing a gui-update is not something I would like to have. That happens quite too easily.
That's why I used a double-tap for the first zoom-step and then one tap for each subsequent step on the watch.

AdrianLxM:
@triplem with one chart this is possible. Just set a flag to the vieport.
Scrolling is a bit bad as swiping to the next fragment and scrolling might get in the way of the other.
Havin both charts zoom and scroll at the same time did not work well for me on my tests.

AdrianLxM:
@triplem it should be possible though: https://github.com/jjoe64/GraphView/issues/155

MilosKozak:
i've never seen time like 20:30

MilosKozak:
btw if you want to swtich to more powerfull graph library we can talk about it

PR #1026, opened 2018-05-07
LP: Don't notify about invalid values during editing.

jotomo:


AdrianLxM:
looks good. There is a notification in the Local-Profile fragment anyway. And an invalid profile cannot be stored.

PR #1031, opened 2018-05-08
Make all prediction colors equally dark

jotomo:
Yellow is confusing since it's used as warning elsewhere and the BG
points are in yellow if above range.

Also personal preference I guess, feel free to close.


jotomo:
Replaced commit to apply @AdrianLxM 's suggestions
screenshot_20180508-213757

PR #1065, opened 2018-06-01
WIP: DanaRS User options view/set

PoweRGbg:


PoweRGbg:
the onSaveClick() is a mess

PR #1066, opened 2018-06-02
TBR emulation

TebbeUbben:
This adds multiple things:
* Update SightParser to compatibilitiy version ``cardea``
* Improved pump limit handling
* Support for emulating TBRs using extended boluses
* Better way to wait for pump responses, pauses thread instead of using a while loop


app/src/main/java/info/nightscout/androidaps/data/Profile.java
```diff
@@ -226,6 +226,10 @@ public synchronized boolean isValid(String from, boolean notify) {
                         basal_v.setValueAt(i, description.basalMinimumRate);
                         if (notify)
                             sendBelowMinimumNotification(from);
+                    } else if (basal_v.valueAt(i) > description.basalMaximumRate) {
+                        basal_v.setValueAt(i, description.basalMaximumRate);
```
AdrianLxM: While the minimum rate is a rather static value depending on the pump type, the maximum value is a dynamically loaded value not necessarily known at that point in time.
Any ideas how to handle that?

TebbeUbben: I set the default value in the pump description and then update it as we have been able to read that value.

app/src/main/java/info/nightscout/androidaps/data/Profile.java
```diff
@@ -243,6 +247,10 @@ protected void sendBelowMinimumNotification(String from) {
         MainApp.bus().post(new EventNewNotification(new Notification(Notification.MINIMAL_BASAL_VALUE_REPLACED,  String.format(MainApp.gs(R.string.minimalbasalvaluereplaced), from), Notification.NORMAL)));
     }
 
+    protected void sendAboveMaximumNotification(String from) {
+        MainApp.bus().post(new EventNewNotification(new Notification(Notification.MINIMAL_BASAL_VALUE_REPLACED,  String.format(MainApp.gs(R.string.maximumbasalvaluereplaced), from), Notification.NORMAL)));
```
AdrianLxM: Should this really be of type `MINIMAL_BASAL_VALUE_REPLACED` ?

TebbeUbben: No, it shouldn't. I'll fix that.

PR #1071, opened 2018-06-03
TabLayout + NavigationDrawer

TebbeUbben:
This adds an improved TabLayout to the main activity and a separate navigation drawer for all plugins that aren't marked to be visible in the configuration generator.

screenshot_androidaps_20180603-232750
screenshot_androidaps_20180603-232754
screenshot_androidaps_20180603-232759


AdrianLxM:
I really like that !!! - apart from some design ideas where we have different taste ;)

1) I would like to change the indicator colour to a neutral one, not yellow

2) Changing `colorAccent` caused the effect that some ok/cancel buttons also got another colour, but not all. Some depend on another component from a library.

3) With DataBindings we could have a setting `compact ActionBar` or similar that  allows for a more compact view like before. It could set L11 to `android:layout_height="30dp"` and add `app:tabMinWidth="0dp" app:tabPadding="0dp"` to the `TabLayout`.

PR #1076, opened 2018-06-04
higher contrast ok and cancel buttons

AdrianLxM:
@swissalpine this should do what you wanted.
Became possible after changing the app theme.


swissalpine:
Yes. THANK you ... I'm waiting now for the sun to give it a try. 👍

PR #1086, opened 2018-06-08
trying to prevent double  values of percentage and timeshift

PoweRGbg:


MilosKozak:
where is this bug comming from?

PR #1097, opened 2018-06-10
Add reservoir view to overview fragment

schmitzn:
screenshot

It's very useful to see the current reservoir level at a glance on the overview page. The background color of the reservoir view changes between green, yellow and red, depending on the remaining units. Thanks to @savek-cc for the idea and the first implementation of this view for combo pumps. I've expanded it to all pumps which provide this information. Therefore I've added ``getReservoirLevel()`` to ``PumpInterface`` and ``reservoirCapacity`` to ``PumpDescription``. I've modified the pump plugins accordingly.


savek-cc:
:+1:

savek-cc:
:+1:

TebbeUbben:
For the Insight the maximum capacity can vary. I suggest replacing that value with a boolean whether the driver supports reading that value since you don't use it anywhere.

app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
```diff
@@ -1192,6 +1194,26 @@ public void onClick(View v) {
...
 
+        if (reservoirView != null) {
+            int reservoirCapacity = pump.isInitialized() ? pump.getPumpDescription().reservoirCapacity : 0;
+            int reservoirLevel = pump.isInitialized() ? (int) Math.round(pump.getReservoirLevel()) : -1;
+            if (reservoirCapacity > 0 && reservoirLevel != -1) {
+                reservoirView.setText(reservoirLevel + " " + MainApp.sResources.getString(R.string.insulin_unit_shortname));
+                if (reservoirLevel < 5) {
+                    reservoirView.setBackgroundColor(Color.RED);
+                    reservoirView.setTextColor(Color.GRAY);
+                } else if (reservoirLevel < 0.25 * reservoirCapacity) {
```
jotomo: IMO "below 25%" isn't not a useful metric. For a 320 U cartridge (Combo) that's 80 U, for a 160 U cartridge (Insight) that's 40 U. To me it matters whether I need to prepare a new cartridge before leaving the house (TDD is 40-60 U). 
I have an alarm at 50 U, which I use to prep and pack a new cartridge.
I think it would be more useful to make warning levels configurable, so the user can adjust it to their use case. That would also remove the need to have the `reservoirCapacity` field in the first place.

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPlugin.java
```diff
@@ -384,6 +386,9 @@ public double getBaseBasalRate() {
         return basalRate;
     }
 
+    @Override
+    public double getReservoirLevel() { return -1; }
```
jotomo: You can look at the `getJSONStatus` method of each pump to see what can be reported, there (line 761), you'll see that the Insight stores the reservoir level in the field `reservoirInUnits`.

jotomo:
Generally (and more so with this commit), I feel the top "ribbon" getting to colorful to be useful, now with 4 colors that both have the same meaning "this is the default value you'll see 80% of the time". It would IMO be better if there was one color for the default value and others indicating a noteworthy state (using colors to attract attention to something non-default).
Something like this (grey=default, white=non-default, though this could be also yellow/red for reservoir levels or a suspended loop):
image

TebbeUbben:
I already suggested the same regarding to the colors, but some others had concerns about it.

swissalpine:
I understand your idea and appreciate the initiative to give it to the community.

But for me it's not right placed. The field in this line are buttons. An info field breaks the logic.
And I'm shure that on my screens isn't enough place for a fourth field ... without cutting the texts of the others.

TebbeUbben:
I recognize them as info fields since they don't do anything when you just click on them.

savek-cc:
We can make it a shotcut to the pump settings ;)

schmitzn:
> I think it would be more useful to make warning levels configurable, so the user can adjust it to their use case. That would also remove the need to have the reservoirCapacity field in the first place.

@jotomo Good point, I agree.


>And I'm shure that on my screens isn't enough place for a fourth field

@swissalpine We could make the visibility configurable as well, so you could simply toggle it off. To fix the logic I'd agree to @savek-cc 's suggestion.

savek-cc:
@swissalpine I originally integrates this into my branch as I needed a quick view (i.e. everytime I look at AAPS) of the remaining reservoir level.
An alarm at 50 units didn't really help (as the combo already has such an alarm) because at 3am I'd just ack and forget. In the morning I'd then leave the house without enough Insulin for the day.
As insulin is even more important for me than a BG level every 5 min, I really wanted to have a prominent placement of that information.

swissalpine:
@savek-cc Everybody has its own use cases and solutions. ;-) But keep in mind that there are different needs out there. This is my solution for my problems (inspired by jotomo):
image

swissalpine:
In my opinion other alerts are not less important and there isn't place for all of them in the button row. ;-)

@TebbeUbben I use the buttons every day several times (Loop status | Profile switch | TT). What goes wrong with your ui?

TebbeUbben:
@swissalpine I use them as well, but 1. they don't look like buttons and 2. they only open a menu when you press them for some time, that's the behaviour of a normal button.

app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java
```diff
@@ -1192,6 +1194,28 @@ public void onClick(View v) {
...
         activeProfileView.setBackgroundColor(Color.GRAY);
 
+        if (reservoirView != null) {
+            boolean showReservoirView = SP.getBoolean(R.string.key_show_reservoirview, false);
+            int reservoirLevel = pump.isInitialized() ? (int) Math.round(pump.getReservoirLevel()) : -1;
+            if (showReservoirView && reservoirLevel != -1) {
+                int levelWarning = SP.getInt(R.string.key_reservoirview_levelwarning, 80);
+                int levelCritical = SP.getInt(R.string.key_reservoirview_levelcritical, 5);
+                reservoirView.setText(reservoirLevel + " " + MainApp.sResources.getString(R.string.insulin_unit_shortname));
+                if (reservoirLevel <= levelWarning) {
```
jotomo: You've mixed up _levelWarning_ here and _levelCritical_ on the _else if_ below :-)

schmitzn:
Unified ribbon colors:

screenshot

AdrianLxM:
As this is not too important information: can't we put it into the pump plugin fragment? It is like this with the Dana pumps and I like it very well there.

savek-cc:
That depends on your definition of "not too important" - for me, it's a must have to see it at a glance whenever I look at AAPS. I usually don't swipe to another screen to get status information and thus missed an almost empty cartrige more than once. So this view keeps me from having to drive home from work for 1h mid-day just because I forgot that the insulin won't last the day (as I acked the alarm while sleeping at 1am).

gregorybel:
yes, I would like to have it on main screen too! where on the main screen is not so important but it should be in main screen. 

Battery's days is important too I think.

And what about using one of the button at the bottom? we could show battery and insulin and when we push on it we jump into the plugin fragment?

swissalpine:
What do you think about this?
It works like a fire light: Dark Grey (nothing to do), Yellow (pay Attention/warning), Red (critical/empty ...)For now it's (as a wrote above: inspired by jotomo) my own solution and not ready for a roll out, but it's very handy for me: I see warnings for Cannula Age, Insulin Age, Reservoir Level, Sensor Age and Battery.
CAGE, IAGE, RES and SEN are ready for all pumps I think, BAT needs some work ...
image

savek-cc:
@swissalpine that would also work for me - even better with sensor age etc.

schmitzn:
@swissalpine This would be fine for me, too. But I think instead of the insulin age we should really take the actual reservoir level. I'd like to integrate your solution into this, if you don't mind providing your source.

AdrianLxM:
One more thing: The top ribbon is already too full for smaller phones. So a screen width detection and different layouts would need to be added.

The solution that Jotomo and Swissalpine (separately?) have would be less problematic (and could be configured by the user).

swissalpine:
I hope I find time to build a branch based on dev with my changes on thursday and I will invate you to help optimizing bat and the configuration of reservoir.

schmitzn:
So this is closed as @swissalpine will create a new branch where we'll continue based on his solution.

CaroGo:
I would love to see this in dev or master version!

nicoleloop:
I also would be thrilled to have this information on the main screen! +1

PR #1101, opened 2018-06-13
setting user options on DanaR 

PoweRGbg:
needs testing


app/src/main/java/info/nightscout/androidaps/plugins/PumpDanaR/comm/MsgSetUserOptions.java
```diff
@@ -55,4 +49,49 @@ public void handleMessage(byte[] bytes) {
         }
     }
 
+    public byte[] getCommByte(int cmd, byte[] data) {
```
MilosKozak: this is not necessary to do here

PR #1102, opened 2018-06-13
New launcher icon

TebbeUbben:
I created a new launcher icon. This is what it looks like:
androidaps icon


MilosKozak:
for nsclient i'd keep NS icon

PR #1108, opened 2018-06-16
Average profile

TebbeUbben:
This adds another profile plugin allowing the user to specify the average values for IC and ISF (like the ones Autotune gives you) and then they will be aligned to the basal curve using the rule of three.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/MainApp.java
+++ b/app/src/main/java/info/nightscout/androidaps/MainApp.java
@@ -51,6 +51,7 @@ import info.nightscout.androidaps.plugins.Overview.OverviewPlugin;
 import info.nightscout.androidaps.plugins.Overview.events.EventNewNotification;
 import info.nightscout.androidaps.plugins.Overview.notifications.Notification;
 import info.nightscout.androidaps.plugins.Persistentnotification.PersistentNotificationPlugin;
+import info.nightscout.androidaps.plugins.ProfileAverage.AverageProfilePlugin;
 import info.nightscout.androidaps.plugins.ProfileLocal.LocalProfilePlugin;
 import info.nightscout.androidaps.plugins.ProfileNS.NSProfilePlugin;
 import info.nightscout.androidaps.plugins.ProfileSimple.SimpleProfilePlugin;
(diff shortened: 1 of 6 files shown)


jotomo:
Given that a change to e.g. the basal at 4:00 effects the outcome an hour later (exact time depending on insulin) and thus basals being set an hour earlier, should the derrived ISF and IC be shifted forward by an hour (or a configurable time)?

philipgo:
Could you explain what the rule of three is?

AdrianLxM:
Rule of three is a shorthand form of cross multiplications. In many languages like German (Dreisatz) or French (Règle de trois) it has become a synonyme of "calculation with proportions" even when you use another method.

philipgo:
I installed this modification and like the idea a lot. However, I would find usability much better if this was an option for the existing profile types (Nightscout, Local Profile) instead of a separate profile type.

LongboatAline:
I like this profile a lot, since I'm a strong believer in consistent parameter definitions, and this profile has the ability to summarily eliminate many sources for errors while metering out someones individual parameters, especially the cases where a given setup works to some extent because those errors that had been made in its evolution are cancelling each other out.
While closing in on a BR setup, this profile ascertains that both ISF and CR are following along as the individual rates get changed, resulting in a rather clear message to the user if the change was a good idea or not (instead of starting to tune another parameter the moment a previous change in BR didn't achieve the desired result)

I've been running this profile for a week now, and while my previous setup hadn't exactly been contradictive in itself, the parameters didn't scale proportionally across the day. (owed also to my attempt to keep autotune happy by simplifying iISF and CR to a single parameter, which I *knew* to be off for a significant part of the day. While this wasn't much of a problem with CR (manual corrections  were easy, and the loop did a good job to compensate for the resulting divergencies), ISF was another matter, leading to a number of persistent lows when the loop tried to battle rescue carbs with a sledgehammer.

This profile definitely helped a *lot* with these situations. (As would a diversified ISF definition have done, but using the BR as a yardstick made it a breeze)

Now's to the problems ;-P
The profile managed to mangle itself, splattering br definitions all over the place and creating gaps in between. How? I didn't notice until I wanted to change the basic CR startvalue. Looks like I'm fine as long as I don't do a profile change, though, so I left it for the time being.
Screenshots: 3 showing the working setuo as of 2018-07-05, and 2 showing the mess in the AP editor six days later... 

##2018-07-05
avgprofile01-ic
avgprofile02-isf
avgprofile03-br

##2018-07-11
avgprofile04-invalid
avgprofile05-invalid

TebbeUbben:
I don't think that this is a problem of the profile itself. It uses the same way to store the basal values as the local profile. The problem must be somewhere deeper.

LongboatAline:
One more thing: Setup Wizard integration.
I just installed a new {patched} rc4 to a tabled with no prior use of AAPS. Selecting average profile in the initial setup wizard didn|t offer a means to edit the average profile, but required the initial profile switch.
Selecting local profile took me to the profile setup first.

LongboatAline:
Today again - BR in the AP editor is shown as 7IE at 21:oo (should have been 0.75), and Target got moved from 90-100 to 0.75-<empty>
The currently active profile is unimpended (95% of the AP due to Up-regulation)
Lets see if/when/ihow it'll fix itself again ;)

swissalpine:
I use this profile since June and I love it! Please, merge it, so more people can use it!

PR #1110, opened 2018-06-17
Use a foreground service for the ongoing notification and enforce using it on Oreo and up

TebbeUbben:
Pretty similiar to #870, makes sure that AndroidAPS won't be terminated by the system.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
@@ -113,7 +113,15 @@ public class ConfigBuilderPlugin extends PluginBase {
         pluginList = MainApp.getPluginsList();
         upgradeSettings();
         loadSettings();
+        setAlwaysEnabledPluginsEnabled();
         MainApp.bus().post(new EventAppInitialized());
+    }
+
+    private void setAlwaysEnabledPluginsEnabled() {
+        for (PluginBase plugin : pluginList) {
+            if (plugin.pluginDescription.alwaysEnabled) plugin.setPluginEnabled(plugin.getType(), true);
+        }
+        storeSettings("setAlwaysEnabledPluginsEnabled");
     }
 
     public void storeSettings(String from) {
--- a/app/src/main/java/info/nightscout/androidaps/plugins/Persistentnotification/DummyService.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/Persistentnotification/DummyService.java
@@ -1,0 +1,28 @@
+package info.nightscout.androidaps.plugins.Persistentnotification;
+
+import android.app.Notification;
+import android.app.Service;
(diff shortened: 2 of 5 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
```diff
@@ -113,9 +113,17 @@ public void initialize() {
         pluginList = MainApp.getPluginsList();
         upgradeSettings();
         loadSettings();
+        setAlwaysEnabledPluginsEnabled();
         MainApp.bus().post(new EventAppInitialized());
     }
 
+    private void setAlwaysEnabledPluginsEnabled() {
```
jotomo: The _OverviewPlugin_ is forcefully enabled without the option to disable it, using _PluginDescription.alwaysEnabled_. That should also work for this plugin?

app/src/main/java/info/nightscout/androidaps/plugins/Persistentnotification/DummyService.java
```diff
@@ -0,0 +1,28 @@
+package info.nightscout.androidaps.plugins.Persistentnotification;
+
+import android.app.Notification;
+import android.app.Service;
+import android.content.Intent;
+import android.os.IBinder;
+import android.support.annotation.Nullable;
+import android.util.Log;
+
+public class DummyService extends Service {
```
jotomo: I think it would be good to add some Javadoc here, as too why this service exists and is force-enabled since the class name is non-descriptive (and tricky to name clearly in this case), to help a reader later on to understand why this class exists at all.

jotomo:
> Pretty similiar to #870

So it superseedes that PR?

app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
TebbeUbben: Always enabled seems to only disable the checkbox in the configuration builder. I added this method to make sure that it is enabled again when updating AndroidAPS or from an Android version before Oreo.

TebbeUbben:
Yes, #870 is redundant with this PR.

app/src/main/java/info/nightscout/androidaps/plugins/ConfigBuilder/ConfigBuilderPlugin.java
jotomo: Good thinking, taking migrations into consideration.

app/src/main/java/info/nightscout/androidaps/plugins/Persistentnotification/DummyService.java
TebbeUbben: Have done that :)

jotomo:
Thanks for the javadoc :+1: 
LGTM

PR #1119, opened 2018-06-20
Treatments tab: add button (if applicable) to delete future treatments.

jotomo:


app/src/main/res/values/strings.xml
```diff
@@ -437,6 +437,7 @@
     <string name="openapsma_scriptdebugdata_label">Script debug</string>
     <string name="openapsama_useautosens">Use AMA autosens feature</string>
     <string name="refresheventsfromnightscout">Refresh events from NS</string>
+    <string name="deletefuturetreatmets">Delete treatments in the future</string>
```
MilosKozak: duplicate?

jotomo: One is the button text, the other is for the confirmation dialog, but since they're the same and unlikely to diverge you're right - removed.

Issue #1090, opened 2018-06-09
Synchronizing Butterknife / Eventbus


AdrianLxM:
This should be done in other fragments as well that use butterknife, I guess.

https://github.com/MilosKozak/AndroidAPS/pull/1085

PR #1120, opened 2018-06-21
Synchronize Butterknife / Eventbus (#1090)

schmitzn:
In reference to #1090 
Adds ``synchronized`` to all fragments that use events together with Butterknife bindings.
The changes in DanaRFragment should be tested by someone else because I don't have a DanaR pump.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/HistoryBrowseActivity.java
+++ b/app/src/main/java/info/nightscout/androidaps/HistoryBrowseActivity.java
@@ -185,6 +185,10 @@ public class HistoryBrowseActivity extends AppCompatActivity {
     }
 
     void updateGUI(String from) {
+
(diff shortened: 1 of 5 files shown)


AdrianLxM:
I believe synchronizing alone won't be enough.

The synchronization will make clear that the unbinding will happen either entirely before or after the synchronized part is called. If it is called before, it can still cause nullpointer-exceptions.

With no having race conditions, it should be enough to check if one of the variables holding a gui element is null. I did so here for example (similar case): https://github.com/MilosKozak/AndroidAPS/pull/1085/files#diff-eab392e708bbebceda44693228490dd8R159

schmitzn:
> If it is called before, it can still cause nullpointer-exceptions.

Good point. I didn't recognize it in your code. Should be fixed now.

AdrianLxM:
Thank you very much!

PR #1121, opened 2018-06-21
TT context menu

jotomo:
Adds a context menu for TT on overview (nicely aligns with the other two context menus). This makes it discoverable how to cancel a TT:
screenshot_1529611588

Dividers/separators around the default TTs would be nice, I'll open an issue for that to do that post-2.0.
(@AdrianLxM groups don't help, needs a workaround or a different approach to creating a menu, so I'm deferring this).


gregorybel:
really cool, i like it

Issue #1103, opened 2018-06-13
Warn the user before a basal profile is written to the pump for the first time


jamorham:
A new user can find that the basal profile of their pump gets overwritten without clear warning in the software that this is going to happen. 

I find this a bit of a dramatic feature and inconsistent with the step-by-step objectives to walk a user slowly in to understanding the app from a safety point of view. For an inept user it is very easy to significantly change insulin dosing via the basal rate in a way which doesn't stop when the app is uninstalled.

In an example situation, a user installs the application, goes through the setup wizard, creates a lazy inaccurate profile for testing using the virtual pump. Later they switch to the real pump to see if communication works at which point the basal profile of the pump is overwritten.

A notification in the overview section appears after this has happened, but by then it has already occurred. 

In some circumstances a user may have tuned their basal profile on their pump through small adjustments over time without noting the details and then that data is lost when it is overwritten by AAPS. More significantly it could happen to a new user who then uninstalls the software and potentially does not realize that the active basal profile on the pump has been changed.

Suggested solution:

Before an operation is done within the app which could cause the pump settings to be updated, the user should receive a yes/no dialog to approve the action. 

Whether this is when the profile is configured or ideally before the data is written to the pump I am not sure the best solution within AAPS architecture, but in english I would suggest a message something like the following:

## ATTENTION:

**AndroidAPS intends to manage the basal profile of any connected pump.**

**The basal profile you have defined within AndroidAPS will be sent to a connected pump and immediately become active. This will overwrite whatever current basal profile is there.  The new daily basal rate is XX.XX U**

**Please confirm that you wish this to happen and understand the consequences.**

**YES / NO**

jamorham:
@AdrianLxM as discussed on gitter

MilosKozak:
👍 thx

AdrianLxM:
Solution here: https://github.com/MilosKozak/AndroidAPS/pull/1168

PR #1135, opened 2018-06-23
Ask to confirm first profile switch

TebbeUbben:
Related to #1103


AdrianLxM:
I believe the case where the profile switch is coming from either Nightscout or the local Profile is missing.

MilosKozak:
it's going to be very hard make use friendly if data comes on background from NS

AdrianLxM:
@MilosKozak it would only be the first time. Basically when setting up AndroidAPS. After that it should never warn again. (Unless the user presses "cancel")

TebbeUbben:
How about showing that dialog on the first time a profile plugin is activated?

schmitzn:
>How about showing that dialog on the first time a profile plugin is activated?

I guess this would be the easiest way for the implementation.

But I think (for the users) it would be better to show the dialog on the first time when the pump profile is actually going to be overwritten. We could use a notification in the overview section with the ability to open the yes/no dialog. So this notification would appear before the basal profile on the pump is being changed. As long as the notification wasn't accepted, the user would have to stay on the unchanged profile.

Maybe it would also be a helpful option to make it possible to import the pumps profile within the setup wizard.

MilosKozak:
OK. let's keep it as it now here. Maybe I'd add link to wiki. what do you think?

AdrianLxM:
The real problem was the profile coming from Nightscout.
And most of the trap we have already eliminated by needing a profile switch.

AdrianLxM:
How about:
 * Virtual Pump as default pump in Config Builder
 * Warning as soon as the user wants to switch to a real pump that this means overwriting the basal settings.

?

jotomo:
Since you user isn't acknowledging a single profile update on the pump, but accepts that AAPS will take control of the pump's basal rate profile hereafter (when and how often it is updated is a detail), asking for confirmation when enabling a pump plugin seems to be a good place.

MilosKozak:
ok, this could be a soulution

AdrianLxM:
Superseded by https://github.com/MilosKozak/AndroidAPS/pull/1168

PR #1137, opened 2018-06-23
Rs sync warn

AdrianLxM:
Warns when the time is synchronized on Dana RS or Rv2 and the time difference was more than 1.5 hours.

This should not disturb on standard DST changes but at least warns the user.
It would still be good to be able to automatically sync the events (if we could get the info off the pump) or erase the pumps "history" that is wrong (missing the commands?). But at least the user is warned when he travels through time zones.

Thanks @JustinWalkerMDPL for reporting this.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1,4 +1,4 @@
-﻿<resources>
+<resources>
     <string name="treatmentssafety_title">Treatments safety</string>
     <string name="treatmentssafety_maxbolus_title">Max allowed bolus [U]</string>
     <string name="treatmentssafety_maxcarbs_title">Max allowed carbs [g]</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1172,6 +1172,8 @@
     <string name="combo_invalid_setup">Invalid pump setup, check the docs and verify that the Quick Info menu is named "QUICK INFO" using the 360 configuration software.</string>
     <string name="custom">Custom</string>
     <string name="key_lockscreen" translatable="false">lockscreen</string>
+    <string name="largetimedifftitle">Large Time Difference</string>
+    <string name="largetimediff">Large time difference: Time in pump was adjuted for more than 1.5 hours. Make sure that reading the history from the pump does not cause unexpected behaviour.</string>
```
MilosKozak: typo "adjuted"

AdrianLxM:
@MilosKozak what do you think about the solution to NOT sync in that case (long time diff) but warn the user to do it manually?

MilosKozak:
this warning comes too late. clean solution would be keep the pump uninitialized until you do it

MilosKozak:
@AdrianLxM

AdrianLxM:
There is still a problem re-initializing the pump after a correct time-sync if the user adjusted the time manually :/

AdrianLxM:
now it gets re-initialized again.

AdrianLxM:
Thank you @Heiner1 for testing the Rv2 part!

MilosKozak:
sounds good to me ne now. thanks guys!

PR #1139, opened 2018-06-23
fixes #1130

### LINKED ISSUE #1130: Nightscout upload only at Wifi
When I set the nightscout upload only with wifi connection and I set a ssid the nightscout client will not work with the log entry "not allowed". When I set the ssid wrapped with "" then it works. Androidaps version rc4.

triplem:
Add StringUtils to allow quotation marks inside of SSIDs.


triplem:
If this is merged, it has to be noted in the release notes, that already existing SSIDs should be stated in the configuration without quotes, otherwise the SSID is not recognized anymore.

PR #1148, opened 2018-06-24
1133

triplem:
Add a unit-test.


AdrianLxM:
@triplem Thanks for the test cases!
They would still work if the Interface would be without implementing functions and have a common abstract class?
That would be the solution I'd prefer.

PR #1149, opened 2018-06-24
Treatments don't have settings.

jotomo:
Before this, treatments preferences opened preferencs of the
oref0 sensitivity plugin.


MilosKozak:
good catch! thanks

PR #1152, opened 2018-06-24
Avoid empty row if no extended bolus is running

swissalpine:


MilosKozak:
looks like there is a conflict. in `shortextmode` it must be `INVISIBLE`

swissalpine:
This was a fix @jotomo introduced weeks ago. And: I use short text mode and this works. With 'invisible' I have an empty row.

swissalpine:
In rc04 you will find the same:
https://github.com/MilosKozak/AndroidAPS/blob/aaps20_rc4/app/src/main/java/info/nightscout/androidaps/plugins/Overview/OverviewFragment.java#L1187
```
if (extendedBolusText.equals(""))
                extendedBolusView.setVisibility(View.GONE);
            else
                extendedBolusView.setVisibility(View.VISIBLE);
```

MilosKozak:
yes, this makes wrong layout in NSCLIENT mode. you need to split it

swissalpine:
Like that?

MilosKozak:
fixed in https://github.com/MilosKozak/AndroidAPS/commit/4b777f8dcc0e7d62d1b3d53b186230033d6fa944

PR #1161, opened 2018-06-26
Use positional format strings in translations

### LINKED ISSUE #1144: [Enhancement] Use positional format strings in translations
To avoid formatted="false" in every translation (which is not supported by default in the Translation Editor of Android Studio) we should use positional format strings in the future.

For details about this, see the following link:
https://www.thedroidsonroids.com/blog/android-strings-xml-tips-tricks

schmitzn:
Fix #1144 
@triplem Could you please take a quick look at it?


MilosKozak:
@AdrianLxM @triplem is it possible to review it somehow automatically

schmitzn:
I've done this by a [quick 'n dirty python script](https://gist.github.com/schmitzn/8441adcfb23d7a775a2f9083684b2a15). In theory, reviewing the script should do it as well as checking all the changes.

PR #1163, opened 2018-06-27
For testing: Wear confirm dialog

AdrianLxM:
@jotomo, @blocklist.de and @gregorybel, could you please test this on a wear 2.0 watch?
It should now have the confirm dialog not as a notification but as an activity with scrolable text.
This should solve the problem that a simple tap can confirm a bolus.

Is the text too large (On my Sony Smartwatch 3 it is rather large but still ok)?

### DIFF
--- a/wear/build.gradle
+++ b/wear/build.gradle
@@ -30,7 +30,7 @@ android {
 
     defaultConfig {
         applicationId "info.nightscout.androidaps"
-        minSdkVersion 21
+        minSdkVersion 23
         targetSdkVersion 23
         versionCode 1
         versionName "1.0.2"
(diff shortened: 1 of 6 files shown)


jotomo:
@blocklist

AdrianLxM:
@jotomo, yes, leftover debugcode.
And thanks @blocklist, will change to a smaller font.

gregorybel:
will do, thank you! but might be not so fast as the other guys.

blocklist:
works for me very good too :-)

AdrianLxM:
Thanks a lot for testing!

PR #1168, opened 2018-06-29
ask user if he switches to a HW pump for the first time

AdrianLxM:
If the user wants to switch to a HW pump for the first time he will get warned that AAPS will take over control of the basal profile. If he presses "cancel" the config builder will jump back to the VirtualPump.

screenshot_20180629-035422

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/interfaces/PluginBase.java
+++ b/app/src/main/java/info/nightscout/androidaps/interfaces/PluginBase.java
@@ -1,11 +1,13 @@
 package info.nightscout.androidaps.interfaces;
 
 import android.os.SystemClock;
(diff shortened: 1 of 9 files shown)


AdrianLxM:
Fixes https://github.com/MilosKozak/AndroidAPS/issues/1103

jotomo:
Nitpicking: don't use the word "synchronizing", as that's not what's happening and is only confusing.
Proposal for the 2nd part of the 1st sentence:
_AnrdoidAPS will copy the basal settings from the profile to the pump, overwriting the existing basal rate stored on the pump._

AdrianLxM:
@jotomo updated, including **AnrdoidAPS** ;)

PR #1107, opened 2018-06-16
Add "status lights" and unify ribbon colors

schmitzn:
As discussed in #1097. The idea is to have reservoir, battery, insulin and catridge status on the homescreen. It works like a fire light: Dark grey (nothing to do), yellow (pay attention/warning), red (critical/empty ...). You can define your individual thresholds in nightscout and in the app preferences.

In my opinion this is very useful, because you see it all at a glance on your homescreen. So this is intended to prevent you from leaving the house without having enough insulin, or missing the battery alarm which came up at night when you were asleep.

Also, I've changed the colors of the top ribbon bar, inspired by @jotomo. So grey is the color for the default value and yellow/red are used to indicate a noteworthy state.

screenshot_20180616-141528

Thanks to @swissalpine as this is based on his sourcecode (and as he gave a lot of good suggestions).

PR #1172, opened 2018-06-29
Updated overview chart

MarkMpn:
Updates the overview chart with:

* All text removed except carbs and boluses
* Tap on a marker to show text in toast popup
* Carbs/boluses shown in Nightscout style with white/blue semicircles proportional to value
* SMBs shown along the BG trend line
* Predictions shown as empty circles
* Tap on a prediction to show which prediction line it is

screenshot_20180629-224810

This solves the following problems for me, hopefully they're useful for others as well:

* School can add long descriptions to notes which become unreadable on the chart
* Caregivers who also follow via Nightscout have to switch between two different layouts
* SMBs are currently shown away from other boluses, giving a misleading impression of how much insulin has been delivered
* Predictions and actual BG values can become confused, and it is not obvious what colours indicate which prediction line


MilosKozak:
I like with some exceptions (my personal opinion):
* SMB is better when not displayed as a bolus. It not readable when displayed as a bolus anyway
* Carbs: previously carbs were displayed in left-down direction and when overlapped with bolus is was still readable
* I'd not stricly follow NS. It's not perfect anyway. We have own colors for insulin and carbs
* Looking forward to comments from designers @TebbeUbben and @alesinka

swissalpine:
I'm not personally asked but I nust write some lines and I excuse the content. And I respect your work!

BUT:
I prefer changing Nightscouts layout not the layout from AAPS. With SMB it's not really readable, the design is in my opinion outdated and in some views not usable while looping (look at the main page, day for day report etc.) . I would like to see a complete redesign of Nightscout not AAPS (a least the SMB triangle)!
Compared to Nightscout the design of AAPS is on a good way. Shurely we can detox some colors (as in PR https://github.com/MilosKozak/AndroidAPS/pull/1107) and optimize some things (as the fantastic new config builder proofs).
But I like what we have allready, please don't change it in this way!!!

AdrianLxM:
Uploaded with Directupload Image Sharer

SMB are not only issued when the BG is rising.
They often take over basal so it really makes sense to have them as indicators with the basal chart.

The amount is not that important IMHO you can have the iob chat below IOB is the important information!

TebbeUbben:
1. We are in the process of improving AndroidAPS' user interface and make it look (consistent with Android / Material Design). Now adding another UI style to AndroidAPS would make things worse. Apart from that Nightscout is pretty far from perfect regarding to that.
2. Displaying SMBs as normal boluses would mess up the chart completely since there isn't enough space for that. It's rather important to know that SMBs have been delivered than how much insulin they represent. Futhermore they are more related to the basal curve, so the should stay there.
> Predictions and actual BG values can become confused, and it is not obvious what colours indicate which prediction line
3.  I never had any problems with that since you only need to remember what color displays which prediction. Confusing the actual BG values with the predictions is also unlikely since there is a dashed lines which seperates the past and the future.
4. I really like the idea to show a marker's value when clicking on it, especially to only notes on demand. This could truly help to cleanup the graph.

MarkMpn:
OK, I've reversed most of the changes except those to only show the notes on things like warnings & BG check entries when tapping on them.

It's good to hear someone is looking at updating the UI already, I'll open a separate issue with some points I hope can be considered as part of that work.

wootmasterslick:
One solution to the overcrowding of the screen would be to adapt the chart rendering so that it is zoomable (in the x axis) with imo ideally 2 fingered pinch move as alternative the option to add +/- buttons would be nice to add for phones wothout multitouch. Another function to add to the chart is to slide left/right to change the time viewed. When zoomed out you can hide detailed decription of certain messages. Upon zoom in messages can become more detailed.

Does this sound like a feasible idea?

haspden:
> One solution to the overcrowding of the screen would be to adapt the chart rendering so that it is zoomable (in the x axis) with imo ideally 2 fingered pinch move as alternative the option to add +/- buttons would be nice to add for phones wothout multitouch. Another function to add to the chart is to slide left/right to change the time viewed. When zoomed out you can hide detailed decription of certain messages. Upon zoom in messages can become more detailed.
> 
> Does this sound like a feasible idea?

I aggress here... pinching & zooming with the X axis would be great... See xDrip+ for an example on how that works & looks or checkout my video here... (please ignore my values I had my Flu jab today :D )

[https://photos.app.goo.gl/KRwF2VtnAxubX7nD7](url)

MilosKozak:
not mergeable anymore

PR #1177, opened 2018-07-01
for review: warning on treatment-id clash

jotomo:
adds parameter 'allowUpdate' to _TreatmentService.addToHistoryTreatmentRecord_.
if _false_ and adding the record fails, raise an alarm.
this covers the cases we know (and those we don't know yet), where timestamp/id clashes lead to lost records (smb at the time of ecarbs, user inputing a record at the time of an existing record, e.g. when backdating hypo carbs, ...).
this is a workaround which stands the chance of being understandable without causing more issues, until  there is time to properly solve the underlying issue of id clashes.

not yet tested, please review if this idea has merit. invocations of the method  from danar* code needs to be checked as to where updates are allowed.

@AdrianLxM

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
@@ -32,7 +32,9 @@ import info.nightscout.androidaps.Constants;
 import info.nightscout.androidaps.MainApp;
 import info.nightscout.androidaps.R;
 import info.nightscout.androidaps.data.Profile;
+import info.nightscout.androidaps.db.BgReading;
 import info.nightscout.androidaps.db.CareportalEvent;
+import info.nightscout.androidaps.db.DatabaseHelper;
 import info.nightscout.androidaps.db.Source;
 import info.nightscout.androidaps.db.TempTarget;
 import info.nightscout.androidaps.interfaces.Constraint;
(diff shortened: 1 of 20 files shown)


jotomo:
@MilosKozak wdyt?

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
```diff
@@ -153,6 +155,13 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         notesLayout.setVisibility(SP.getBoolean(R.string.key_show_notes_entry_dialogs, false) ? View.VISIBLE : View.GONE);
         notesEdit = view.findViewById(R.id.newcarbs_notes);
 
+        BgReading bgReading = DatabaseHelper.actualBg();
+        if (bgReading != null && bgReading.value < 72) {
+            startHypoTTCheckbox.setOnCheckedChangeListener(null);
+            startHypoTTCheckbox.setChecked(true);
+            startHypoTTCheckbox.setOnClickListener(this);
```
AdrianLxM: in other places when toggling the checkbox you afterwards do
`startHypoTTCheckbox.setOnCheckedChangeListener(this);`, not `startHypoTTCheckbox.setOnClickListener(this);`.
Is that the correct behaviour?

AdrianLxM:
`TreatmentsPlugin.getPlugin().addToHistoryTreatment(detailedBolusInfo, false);` seems to be always called with `false`, never with `true`. Is that correct?

app/src/main/java/info/nightscout/androidaps/plugins/Overview/Dialogs/NewCarbsDialog.java
jotomo: Unchecking a radio item must be done via a click-handler, since no _CheckedChanged_ is triggered, since deselecting a radio is not normally possibly other than selecting a different radio.
i've simplified the logic a biit and referenced the commit explaining this

jotomo:
Yes, the argument is currently always _false_. the combo initially needed to set it to _true_, when reading history, but i've updated that to be more explicit (to show the intend more clearly). the same may be necessary for danar* pumps, so i left the option in, if those pumps need it. please check the calls if the arguments needs to be _true_ anywhere. if not, it can be removed, but it needs to be properly reviewed beforehand (and lacking a danar* i can't test this)

AdrianLxM:
With the DanaRS this will cause a lot of alarms every time the history is read. Still needs some work

AdrianLxM:
The version as is did run very well so far. Through half a day including 2 main meals and many eCarbs and SMB

PR #1183, opened 2018-07-03
Autosens fix

### LINKED ISSUE #1182: Autosens shows "Autosens deactivated"
After switching to the OpenAPS Oref1 Sensitivity Plugin, the sensitivity is shown shortly in the OpenAPS SMB Tab, but is then shown as "Autosens deactivated".

triplem:
Should fix #1182. 

@jotomo could you please review it shortly?

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/interfaces/SensitivityInterface.java
+++ b/app/src/main/java/info/nightscout/androidaps/interfaces/SensitivityInterface.java
@@ -1,18 +1,6 @@
 package info.nightscout.androidaps.interfaces;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import java.util.Arrays;
-import java.util.Date;
-import java.util.List;
-
-import info.nightscout.androidaps.Config;
-import info.nightscout.androidaps.R;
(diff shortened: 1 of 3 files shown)


triplem:
Revert the last change, and now try to call the latest AutosensData for the "current" variable. Seems to work now. Could be, that this bug is also in the oref0 plugin, but this is not fixed.

MilosKozak:
so the only change now is to call it synchronized? https://github.com/MilosKozak/AndroidAPS/pull/1183/files#diff-caca8d8281d898e69d67edeb3651482cR74
@triplem

triplem:
The change is really the call to getLastAutosensDataSynchronized instead of getAutosensData(time). This delivers the latest one, whereas the previous call returned null, because it is looking for the one based on the given time.

typ1-brods:
Same behaviour when using Oref0, average or AAPS algorythm. Showing 'Autosens deactivated' in script, but in settings it is enabled. It is working one time after changing the algorythm, then 'deactivated'

MilosKozak:
for me it's definitely better. merging

MilosKozak:
hmmm it was wrong. you call detection to cuurent time not to `toTime`

PR #1184, opened 2018-07-04
Idea: Fabric-log used openaps version upon invocation.

jotomo:
To get an idea when older plugins can be sensibly removed.


triplem:
Shouldn't we do this in all Plugins? That is a lot of messages but could clarify how the application is really used.

jotomo:
Superseeded by https://github.com/MilosKozak/AndroidAPS/pull/1204

Issue #1186, opened 2018-07-04
[Enhancement] Adding an Alarm when setting Carb time


harispoljo:
When pre bolusing it would be great if an alarm would trigger(optional) after carb time has passed in case you forget to eat.

PR #1189, opened 2018-07-07
Added alarm function when using carbtime

harispoljo:
Implemeted it mostly because I want to get familiar with the project and how it works with contributing

Related Issue: #1186


AdrianLxM:
Hi @harispoljo! Thank you very much for this PR.

Just a small warning: We do hae a feature freeze to eventually be able to release AAPS 2.0. It therefore might happen that we don't look at this PR until we start working on AAPS 2.1.

In case that happens, it is NOT that we don't appreciate you helping, it is our limited resources and the strict focus on a stable version.

Feel free to contact me any time via gitter.

triplem:
One note, shouldn't the Checkbox label be translateable? The eat_reminder_notification could use a positional argument.

Thanks for this PR.

PR #1199, opened 2018-07-09
Use pump bolus increment for SMB rounding

MarkMpn:
Replaces the current 0.1U step for SMBs with the minimum bolus size allowed by the pump, so a DANA RS could deliver a 0.05U SMB.

This is **untested** - any advice on how to set up tests for this logic would be appreciated!

Apologies if the preference is for changes such as this to be made in oref0 first and then ported to AAPS. If that is the case then is there anyone who could help me out with making those changes?


MilosKozak:
we are trying to keep determinebasal unchanged to have the possibility of easier upgrade. propose the change to oref1. I agree with this.
Closing for now

PR #1202, opened 2018-07-09
Adding text about  temp target to objective 2

MarieT1D:


AdrianLxM:
Could you please add the same for all other languages that have a translation of that string?

You can add that in English - it makes sure the text is there. If people see it in english they will translate it. If they don't see it at all, they won't.

PR #1204, opened 2018-07-09
Fabric-log enabled plugins once a day.

jotomo:


jotomo:
Like so? (New commit: https://github.com/MilosKozak/AndroidAPS/pull/1204/commits/09a1d23bc89bd4cf5f9ebe7cb2bfdc94fdd3211a)

MilosKozak:
no, the same principle like plugins. send once a day not at app start because some apps are not restarted at least once a day.
The name of the event can be changed when it will not be app start anymore

MilosKozak:
👍 THANKS!

Issue #1196, opened 2018-07-08
INSIGHT: Bolus calculator - can't add + correction by click on plus symbol


maskac:
When using Insight plugin, can't use "+" icon in calculator for insulin correction. When i "click" + , message "Bolus constrains applied" is show. I can entry any number manual.
![img_1848](https://user-images.githubusercontent.com/16052233/42422543-90009728-82e8-11e8-9068-72a24a123f8c.JPG)
[AndroidAPS.log](https://github.com/MilosKozak/AndroidAPS/files/2173963/AndroidAPS.log)
Tested on version 2.0RC4 and 1.60e-dev (today builded (8.7.2018)

JimmyFoks:
Similar problem with Insight and AAPS version 2.2.2. Unable to add over plus button - and manually entered correction +0.02 - +0.04. Higher values (0.05+) can be entered.

MilosKozak:
@TebbeUbben

PR #1209, opened 2018-07-11
Fixed problem with Corr numberpicker when using Insightpump

harispoljo:
Can @jamorham check?

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPlugin.java
@@ -914,6 +914,9 @@ public class InsightPlugin extends PluginBase implements PumpInterface, Constrai
     public Constraint<Double> applyBolusConstraints(Constraint<Double> insulin) {
         if (statusResult != null) {
             insulin.setIfSmaller(statusResult.maximumBolusAmount, String.format(MainApp.gs(R.string.limitingbolus), statusResult.maximumBolusAmount, MainApp.gs(R.string.pumplimit)), this);
+
+            if(insulin.value() != 0)
+                insulin.setIfGreater(statusResult.minimumBolusAmount, String.format(MainApp.gs(R.string.limitingbolus), statusResult.maximumBolusAmount, MainApp.gs(R.string.pumplimit)), this);
         }
         return insulin;
     }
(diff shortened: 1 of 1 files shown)


harispoljo:
Fixed #1196

MilosKozak:
@TebbeUbben ok?

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPlugin.java
```diff
@@ -914,7 +914,6 @@ private PumpEnactResult pumpEnactFailure() {
     public Constraint<Double> applyBolusConstraints(Constraint<Double> insulin) {
         if (statusResult != null) {
             insulin.setIfSmaller(statusResult.maximumBolusAmount, String.format(MainApp.gs(R.string.limitingbolus), statusResult.maximumBolusAmount, MainApp.gs(R.string.pumplimit)), this);
-            insulin.setIfGreater(statusResult.minimumBolusAmount, String.format(MainApp.gs(R.string.limitingbolus), statusResult.maximumBolusAmount, MainApp.gs(R.string.pumplimit)), this);
```
TebbeUbben: You should leave that line, but skip it if the insulin amount is 0.

harispoljo: I booted it up on my rig and it still didn't work however I noticed that it only trigger for values between 0 and 0.04

app/src/main/java/info/nightscout/androidaps/plugins/PumpInsight/InsightPlugin.java
```diff
@@ -914,7 +914,9 @@ private PumpEnactResult pumpEnactFailure() {
...
+
+            if(insulin.value() < 0 && insulin.value() > 0.04 )
```
jotomo: This condition is always false. _value_ can't be negative AND greater 0.04 at the same time.
If this is to exclude an amount the pump can't deliver, then I think that's worth a comment or constant to help the maintainer of the code understand why this is here.

harispoljo:
Wait why did this get merged?

The problem with the numberpicker is that it increaments with 0.01 so only way to get past 0.04 is to 
type in the value 0.05 instead of pressing + because that would trigger "Bolus constrains applied". I dont understand why there is a check for the numberpicker when there always will be a check if the amount of insulin after you click ok.

TebbeUbben:
Okay, then you need to fix the number picker. The insight plugin is fine now. What you have fixed is the issue that AAPS wanted to deliver 0.05 even if you didn't enter any insulin in the treatment dialog. Trying to deliver less than the minimum amount (which is dynamically read from the pump) will cause it to throw an error.

PR #1224, opened 2018-07-17
Save & restore dialog state on screen rotate

PoweRGbg:


MilosKozak:
YES! more like this please

PR #1226, opened 2018-07-17
Disable history browser unless on dev-branch.

jotomo:
If history browser is part of a release, users will expect it to be fully working. Since the history browser isn't there yet, I think it should be restricted to the `dev` branch until it's further along to avoid bug reports about missing functionality.


MilosKozak:
agree. thanks @jotomo

PR #1227, opened 2018-07-18
fix for carbs and wizard dialogs on screen rotation

PoweRGbg:


MilosKozak:
checkboxes are not restored

PoweRGbg:
On my emulator they work. Do you have problems with the checkboxes in NewInsulinDialog also ?

MilosKozak:
in emulator it works. wondering why it's not necessary to restore checkboxes and radiobutton state but you have restore TextEdits
any idea @jotomo @AdrianLxM ?

MilosKozak:
hmm looks like there is default `android:saveEnabled = "true"`
then you don't need to save state right @PoweRGbg ?

jotomo:
Default controls restore state for me. @MilosKozak We're not directly using _TextView_, but _NumberPicker_. We may need to extend _NumberPicker_ to restore the state of its children.

MilosKozak:
@jotomo do you know how to do it without changing all dialogs. we must be able to disable this behavior for dynamicaly generated dialogs

jotomo:
Hm, https://github.com/jotomo/AndroidAPS/commit/8f871b362057078a2eb4f5f4c457b9e841403bbd

MilosKozak:
is it part of some PR?

jotomo:
Nope, just a quick commit to see if it works (yes) and how we like this approach :-)

PR #1228, opened 2018-07-18
[2.1] [WIP] Fix premature bolus progress dismissal with SMB

### LINKED ISSUE #1218: Parallel SMB visually overrides manual bolus
I am using AndroidAPS 2.0 RC4
Build ace89f7d8
With Combo Plugin

When submitting a bolus of approx 4 insulin units using the calculator, I wondered why the popup with the progress bar indicating the delivery of the bolus only showed a delivery of 0.1 units and then disappeared.

When checking the Combo Tab, it showed that the delivery of the 4 iu bolus was still in the queue and I could feel by the pumps vibration that the bolus was delivered. However, the progress bar did not appear again.
Checking the SMB tab revealed, that an SMB of 0.1 had just been submitted with the last arrival of a blood glucose value.

I guess the progression bar popup somehow showed the SMB delivery an the bolus delivery was not indicated.
While this is not a critical issue, as insulin dosage was correct, it can confuse users and maybe make them re-deliver the larger bolus, as they only saw the super micro bolus was delivered.

Unfortunately, I do not have any logs.

jotomo:
Fixes #1218. Most likely to occur with Combo due to its slowness.

Tested only with Virtual Pump so far.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/queue/QueueThread.java
+++ b/app/src/main/java/info/nightscout/androidaps/queue/QueueThread.java
@@ -62,7 +62,6 @@ public class QueueThread extends Thread {
                 long secondsElapsed = (System.currentTimeMillis() - connectionStartTime) / 1000;
 
                 if (!pump.isConnected() && secondsElapsed > Constants.PUMP_MAX_CONNECTION_TIME_IN_SECONDS) {
-                    MainApp.bus().post(new EventDismissBolusprogressIfRunning(null));
                     MainApp.bus().post(new EventPumpStatusChanged(MainApp.gs(R.string.connectiontimedout)));
                     log.debug("QUEUE: timed out");
                     pump.stopConnecting();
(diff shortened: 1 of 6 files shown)


app/src/main/java/info/nightscout/androidaps/queue/QueueThread.java
```diff
@@ -62,7 +62,6 @@ public final void run() {
                 long secondsElapsed = (System.currentTimeMillis() - connectionStartTime) / 1000;
 
                 if (!pump.isConnected() && secondsElapsed > Constants.PUMP_MAX_CONNECTION_TIME_IN_SECONDS) {
-                    MainApp.bus().post(new EventDismissBolusprogressIfRunning(null));
```
AdrianLxM: How will the dialog be closed in case the connection to the pump cannot be established and therefore no `execute()` is called?

jotomo: Good point, in that case the dialog should indeed be closed, no matter which kind of bolus(es) were requested.
However, shouldn't the first block (line 65ff) be moved into the "not triggering BT watchdog" branch (or at least the dialog dismiss?)

jotomo: Reverted the bogus change

jotomo:
This is crap and most certaintly creating other issues, closing.

PR #1229, opened 2018-07-18
[2.1] [WIP] Make NumberPicker persist value; TransientNumberPicker where not desired

jotomo:


MilosKozak:
@jotomo it doesn't work

PoweRGbg:
NewNSTreatment work is done in my #1227

jotomo:
@PoweRGbg The approach evaluated here is to use the standard mechanism from Android, so that Android does most of the work, so we needn't persist each field manually (and then forget to do so when adding new dialogs). With that approach NewNSTreatment wouldn't need to persist and restore inputs. It would require changes to make sure input values aren't overriden when they are already set by Android.

PoweRGbg:
@jotomo I see and understand what you've done, but BG and tempTarget values are overridden with every new BG value, or default temp (as you commented in #1227), and in my version, they are set back to the previous savedInstanceState ( not done by Android ).

AdrianLxM:
Having the BG and TT being changed with a new BG value is a cool feature. @RoumenGeorgiev did you manage to do that?

PoweRGbg:
@AdrianLxM  it done already in the NewNsTratment.class, but not by me and it gets the latest BG or default temp, but it overrides the one from user input ( if there is one )

Issue #1242, opened 2018-07-22
AAPS crahes when tap on run in SMB


JulienBub:
- RC5-1
- not tested with AM/ AMA yet

PR #1243, opened 2018-07-22
Abort APS run if AS requested but not yet available.

jotomo:
Should only occur during startup or when something else is broken
already, so just return.


jotomo:
Fixes #1242

PR #1253, opened 2018-07-24
NewCarbsDialog: correct situation where Hypo TT can't be unchecked.

jotomo:
Tested, but please review and test before merging.


swissalpine:
Works for me. Thanks for the quick fix.

PR #1290, opened 2018-08-01
Autotune compatibility change

MarkMpn:
Includes the `rate` property in NS temp basal treatment records that is expected by Autotune.

For percentage temp basals, the rate is calculated based on the current profile rate. This is imperfect when the temp basal overlaps a change in the underlying basal rate, but should generally be close enough to get useful results from Autotune.


app/src/main/java/info/nightscout/androidaps/plugins/NSClientInternal/NSUpload.java
```diff
@@ -81,12 +82,16 @@ public static void uploadTempBasalStartPercent(TemporaryBasal temporaryBasal) {
         try {
             SharedPreferences SP = PreferenceManager.getDefaultSharedPreferences(MainApp.instance().getApplicationContext());
             boolean useAbsolute = SP.getBoolean("ns_sync_use_absolute", false);
+            Profile profile = ProfileFunctions.getInstance().getProfile();
```
MilosKozak: it should be `getProfile(temporaryBasal.date)`

MilosKozak:
👍

PR #1322, opened 2018-08-09
[WIP] Add Possibility to send Preferences as well

triplem:
Right now, the preferences are exported, without the user being asked. This is to not have too much interaction with the user.


AdrianLxM:
The preferences do contain the key for the pump control and the session (Dana RS) and the API-Secret for Nightscout.
I feel uncomfortable with users sending me that kind of data.

triplem:
Uh, that is correct. I will remove those keys before sending.

PR #1335, opened 2018-08-14
Additional calculator safety limits

MarkMpn:
Not sure if this is generally useful or not, but creating a PR for it just in case.

We find that at many times, only a portion of the total bolus for carbs is needed up front. Giving all the bolus as calculated by `carbs / ic` quickly leads to a hypo, while giving a limited amount initially and recording all the carbs allows SMBs to deliver the remaining insulin safely over time. While we can account for this and deliver the insulin manually, other caregivers (school etc.) tend to use the calculator.

This PR adds a new configurable cap to the insulin that the calculator will suggest for the new carbs that are being entered. Leaving it set to the default 0 means that the current behaviour is left unchanged, while any other value will be used as a limit for the Carbs element of the calculator.


MarkMpn:
The build for this is failing, reporting an error with the Gradle version expecting to be 4.6. Updating the Gradle version on my machine does seem to work, but not sure if this is a valid fix or if it's something related to my code changes.

I've also added some changes to get the calculator to not recommend a bolus if the BG is below a defined threshold. I could do with some help here to tidy this up however:

1. The preferences page requires a value between 0 and 5, on the assumption that you're entering the value in mmol. I'm not sure how to change the settings on the input validation to allow different values if your profile is in mg/dl
2. The calculator tests fail due to a dependency on `SP`, which fails on `static SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(MainApp.instance().getApplicationContext());`, presumably because there is no instance of `MainApp` when the tests are being run. I'm not familiar with this testing framework - is it possible to inject a different context for testing, or should I refactor the code to pass through the settings as additional parameters instead of trying to read the settings directly in the calculator code?

Thanks!

MarkMpn:
I've just pushed some additional changes to this, so the new calculator safety features introduced by this now are:

* Calculator does not produce any recommended bolus if the current BG is below a threshold (prevents making hypos worse in case of entering carbs through the calculator)
* Limits insulin recommended for new carbs (where the full amount of insulin would often cause a hypo due to the insulin acting quicker than the carbs, this gives an initial bolus at a safer level and allows the OAPS algorithm to make up the rest as needed later on)
* Limits insulin recommended for existing COB to match the current IOB (allows IOB and COB options to be left turned on without the risk of the calculator recommending large boluses due to high COB, while still recommending boluses for high BG)
* Limits total IOB that would be generated after delivering the recommended bolus

All these are options configured in the safety preferences section, and the default behaviour is unchanged.

MilosKozak:
not mergeable anymore

PR #1343, opened 2018-08-18
Insightfix

bertroode:
I tested the fix from Milos Kozak and concluded that pump history records where overwritten when the was a intent for NightScout. To block this some extra code is added.


bertroode:
Made a mistake somewhere, will retry

PR #1346, opened 2018-08-18
Revison of insightfix because of overwritten pump records in the hist…

bertroode:
…ry file and duplicate values in NS


bertroode:
Added to if statement to the insightfix to prevent overwrting of records in the history file and duplicate values in NS

PR #1368, opened 2018-08-27
POC for file reading

triplem:
There is a new Plugin (SourceFilePlugin in the Source package), which is now called in the Maintenance-Plugin via a new Button (Read Source File). This plugin just reads a specific file in the raw folder (testfile.json) and prints out all lines in the log. Nothing fancy.


MilosKozak:
how do i run this from windows/command line?

triplem:
ah, that was the question. this is a misunderstanding on my side. will work on this one.

Issue #1442, opened 2018-09-16
[Enhancement]: Translate Unit in the wizard total field?


quizzmaster:
Maybe the Unit which is shown in the wizards total field should be translated to match local diffences?
For example: In Germany "IE" instead of "U" is used as Unit for Insulin.

PR #1444, opened 2018-09-16
Using localization and SI convention for displaying units

quizzmaster:
At the moment "U" is used for all languages, except for some points in the GUI.

This PR provides
- the localized insulin unit short name as unit (#1442)
- a space before the units according to the SI convention

I'd like to read the meanings of others about this topic. Maybe someone could also check if this PR matches all occurrences of units.

The spaces before the units within the graph view might let the text get too big. What's your opinion?


quizzmaster:
The work I've done isn't wasted, because the occurrences of the units can be located much better. My next step will be the format strings.

I'm sorry for not waiting for an answer before I started. Only wanted to try this out and share it.

MilosKozak:
do a small sample first

quizzmaster:
@MilosKozak I did a small example for the APSResult within the last commit. Here are the snippets I made as an example:
```
<string name="format_rate_percent"><![CDATA[<b>Rate</b>: %1d%% (%2$.2f U/h)]]></string>
<string name="format_rate_units"><![CDATA[<b>Rate</b>: %1$.2f U/h (%2$i%%)]]></string>
<string name="format_duration"><![CDATA[<b>Duration</b>: %1$d min]]></string>
```
```
else if (usePercent)
    ret = MainApp.gs(R.string.format_rate_percent, percent, percent * pump.getBaseBasalRate() / 100d) + "<br>" +
          MainApp.gs(R.string.format_duration, duration) + "<br>";
else
    ret = MainApp.gs(R.string.format_rate_units, rate, rate / pump.getBaseBasalRate() * 100d) + "<br>" +
          MainApp.gs(R.string.format_duration, duration) + "<br>";
```
Maybe additional format strings are necessary for the string version of the APSResult.

My suggestions are format strings for:
- insulin with unit ("%1$.2f U")
- basal rate with unit ("%1$.2f U/h")
- aps result, spanned (HTML)
- aps result, string
- maybe some cases I'm just not thinking about

The reason isn't translatable, cause of the usage of a javascript asset, correct?

quizzmaster:
> @MilosKozak I did a small example for the APSResult within the last commit.

@MilosKozak What about this?

app/src/main/assets/OpenAPSMA/determine-basal.js
```diff
@@ -144,7 +144,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     }
     if (eventualBG < profile.min_bg) { // if eventual BG is below target:
         if (rT.mealAssist.indexOf("On") == 0) {
-            rT.reason = "Meal assist: " + meal_data.carbs + "g, " + meal_data.boluses + "U";
+            rT.reason = "Meal assist: " + meal_data.carbs + " g, " + meal_data.boluses + "U";
```
AdrianLxM: This is determine basal and should be fixed upstream?

app/src/main/java/info/nightscout/androidaps/plugins/Loop/APSResult.java
```diff
@@ -124,20 +124,19 @@ public Spanned toSpanned() {
...
+                      MainApp.gs(R.string.format_duration, duration) + "<br>";
 
             // smb
             if (smb != 0)
-                ret += ("<b>" + "SMB" + "</b>: " + DecimalFormatter.toPumpSupportedBolus(smb) + " U<br>");
+                ret += ("<b>" + "SMB" + "</b>: " + DecimalFormatter.toPumpSupportedBolus(smb) + " " + MainApp.gs(R.string.insulin_unit_shortname) + "<br>");
 
             // reason
             ret += "<b>" + MainApp.gs(R.string.reason) + "</b>: " + reason.replace("<", "&lt;").replace(">", "&gt;");
+            log.debug("quizzi: " + ret);
```
AdrianLxM: personal logging

AdrianLxM:
It might be that the whole "U/h" should be externalized. I guess that in bulgarian or chinese the hour is abbreviated differently as well.

I know that the space between number and unit is encouraged... but please check with different phone sizes that this does not move the unit to a different line.

MilosKozak:
it's unmergeable anymore. But should be definetely done. PM us if you want to do it with format strings again

Issue #1501, opened 2018-10-12
Aaps freeze and can't be started through database corruption


osodebailar:
Occured with dev form 07.10.2018
and occured to other users in looped-de group.

If you clicked on aaps icon, the app screen is grey and after a while a system message box appears:
'Android APS regiert nicht mehr' 

No chance to start the app anymore until i updated with an older dev from 05.10.2018
The App but the loop don't run.
If you click manual on the button in the loop tab or smb tab to recalculate you see that calculation can't run without autosense data
Additional in the main screen the COB: entry shows 'N/A'
I thin the database is corrupted.
The only thing you can do to heal the situation is a database reset.
After reseting the database the app runs fine
![screenshot_20181012-061623](https://user-images.githubusercontent.com/25795894/46858120-dac2e280-ce0a-11e8-96f7-602767e1a487.png)
![screenshot_20181012-064812](https://user-images.githubusercontent.com/25795894/46858121-dac2e280-ce0a-11e8-8a77-a7a632dabbe0.png)
![screenshot_20181012-074541](https://user-images.githubusercontent.com/25795894/46858122-dac2e280-ce0a-11e8-9cda-8b9ca219a8f9.png)
[AndroidAPS._2018-10-11_20-19-53_.10.zip](https://github.com/MilosKozak/AndroidAPS/files/2472283/AndroidAPS._2018-10-11_20-19-53_.10.zip)
![img_2292](https://user-images.githubusercontent.com/25795894/46858123-db5b7900-ce0a-11e8-8e67-e092988b1603.JPG)
![screenshot_20181012-024222](https://user-images.githubusercontent.com/25795894/46858124-db5b7900-ce0a-11e8-94c4-5567c1bf1d15.png)

osodebailar:
One log
[AndroidAPS._2018-10-11_20-19-53_.10.zip](https://github.com/MilosKozak/AndroidAPS/files/2472290/AndroidAPS._2018-10-11_20-19-53_.10.zip)

MilosKozak:
fixed by https://github.com/MilosKozak/AndroidAPS/commit/20fa8ed371ae301f33538d5009e35fb54158fdb4
thanks for reporting

PR #1502, opened 2018-10-12
[For review] Correct off-by-one error calculating average.

jotomo:
Stumbled across this, might help with #1501.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/IobCobCalculator/IobCobCalculatorPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/IobCobCalculator/IobCobCalculatorPlugin.java
@@ -146,7 +146,7 @@ public class IobCobCalculatorPlugin extends PluginBase {
                     return false;
                 }
             }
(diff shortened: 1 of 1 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/IobCobCalculator/IobCobCalculatorPlugin.java
```diff
@@ -146,7 +146,7 @@ public boolean isAbout5minData() {
                     return false;
                 }
             }
-            double averageDiff = totalDiff / (bgReadings.size() - 1) / 1000d;
+            long averageDiff = totalDiff / bgReadings.size() / 1000;
```
MilosKozak: please explain the idea to have result long

jotomo: A precision smaller than one second isn't really needed. It's a minor point though, feel free to just take over the -1 bit.

PR #1503, opened 2018-10-12
[For review] Queue: correct removal of queued command by type.

jotomo:


MilosKozak:
it removed only one half, right?

jotomo:
Yes, if two commands of the same type were next to each other in the queue, the second one wouldn't be checked. Currently, that shouldn't happen (since a later command of the same type would replace a queued one), but if that ever changes, this would cause a headache of a bug.

MilosKozak:
@jotomo thanks. bow.

PR #1505, opened 2018-10-14
replace swedish minus sign with default minus sign

AdrianLxM:


Pallesauka:
With this fix I can set negative correction values using the calculator and Swedish language in AAPS.

PR #1518, opened 2018-10-25
UAMSMBBasal Minutes and Min SMB dosage

Tornado-Tim:
This PR adds some features from oref1 master to AAPS. (nothing in this alters with oref behavior from master, just brings things up to date)

First Feature: 
maxUAMSMBBasalMinutes: This allows the system to give a more hefty smb only when UAM is detected.
By default this is set to the same amount as maxSMBBasalMinutes.

Second Feature:
A configurable min value for SMB  amount to be increased by, by default this is set to 0.1 units, but gives the option to go lower than that all the way down to 0.05units (limited to 0.05 to 1 unit), extremely usefull for people who are very sensitive to insulin but still want to enable SMB.

This PR requires translations for languages other than English....

This will be my first of many PR's to AAPS!

Cheers @MilosKozak and @AdrianLxM

### DIFF
--- a/app/src/main/assets/OpenAPSSMB/determine-basal.js
+++ b/app/src/main/assets/OpenAPSSMB/determine-basal.js
@@ -995,8 +995,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
 
     // eventual BG is at/above target
     // if iob is over max, just cancel any temps
-    // if we're not here because of SMB, eventual BG is at/above target
-    if (! (microBolusAllowed && rT.COB)) {
+    if ( eventualBG >= max_bg ) {
         rT.reason += "Eventual BG " + convert_bg(eventualBG, profile) + " >= " +  convert_bg(max_bg, profile) + ", ";
     }
     if (iob_data.iob > max_iob) {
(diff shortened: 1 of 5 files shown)


MarkMpn:
My intention with the `bolus_increment` setting was to take it automatically from the pump configuration rather than having the user have to enter it manually - see #1199

Tornado-Tim:
> My intention with the `bolus_increment` setting was to take it automatically from the pump configuration rather than having the user have to enter it manually - see #1199

In principle I agree, but I decided it should be completely agnostic to a pump setting as it should be looping option not a pump option.
Also the Pump setting would be locking the increment to what the pump has (0.05 and 0.1 on a Dana RS and R), by having a manual setting you allow even more adjustment (0.05 to 1 unit), which follows the stock OpenAPS oref1 design.

A constraint should be applied though to ensure the min smb amount can not be lower than the bolus increment on the pump.

I will add this into the PR.......

AdrianLxM:
Can we close this as https://github.com/MilosKozak/AndroidAPS/pull/1522 has a newer version of determine-basal.js already included?

Tornado-Tim:
No don't close as this contains 2 features which are not in the other PR. I decided to take this approach as these features are in oref1 0.6.2 and #1522 contains features from oref dev (to be 0.7.0)

PR #1521, opened 2018-10-26
Smsrefactor

MilosKozak:


MilosKozak:
Still adding tests but is this OK for you?
@AdrianLxM @TebbeUbben

PR #1522, opened 2018-10-26
SMB Interval addition

Tornado-Tim:
Adds another SMB feature which is current upstream versions of oref1.

The SMBInterval Parameter can restrict how frequently SMB's can be dosed, limits apply from 1 min to 10.
This can also ensure that SMB's will be able to delivered more frequently if required.

The default is 3 min which has not changed.

### DIFF
--- a/app/src/main/assets/OpenAPSSMB/determine-basal.js
+++ b/app/src/main/assets/OpenAPSSMB/determine-basal.js
@@ -1085,17 +1085,23 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
             }
             rT.reason += ". ";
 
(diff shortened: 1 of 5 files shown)


MilosKozak:
we were alway trying to not touch determine basal because it makes it lot of easier to update to new version. if you think there are some changes needed create PR to OAPS and we'll take as a new file for new verison

Tornado-Tim:
Hi @MilosKozak, no external changes made to determine basal..... this only coming from upstream.

I am just PR'ing individual features added to determine_basal in order to integrate a bit more cleanly for integration purposes.

This Feature is introduced in OpenAPS 0.7.0.....

PR #1541, opened 2018-11-05
(WIP) First Implementation of a profile copying (from NS to Local)

triplem:
This implements copying a profile from NS to the local profile. Right now, it just copies the active profile.


MilosKozak:
currently selected profile is hard to change. would be possible to select from list of available NS profiles?

triplem:
Currently, i am trying to fix a couple of bugs on this one, but the button moved to the NS Profile, where it is shown next to the "Activate Profile" Button. The curently selected Profile is then copied. Guess that this will solve the above mentioned issue, no?

MilosKozak:
ok. yes. it should be enough. 
i'm only not sure if getCurrentProfile() returns the one selected

PoweRGbg:
@triplem  I just moved my code to the maintenance plugin and you moved it :) I will create a PR with the uploading LocalProfile to NS and you decide where to put it or comment here

PoweRGbg:
I'm creating a new profile named "LocalProfile".

triplem:
The logic is now fully integrated in the NSProfile Plugin. There is a button on the NS profile fragment now, which copies the currently selected profile to local.

triplem:
@MilosKozak Me is using getSelectedItem() now, instead of the currentProfile. This makes it possible to select one profile and copy it to local, even if it is not active currently.

triplem:
@PoweRGbg What happens, if the profile is existing already. Do you overwrite it?

PoweRGbg:
@triplem It overwrites it, but still there is a "validFrom" field and it's appearing as a dropdown menu in NS Profile editor

triplem:
I try to add the NS Upload as well. Thanks to @PoweRGbg, but help is really appreciated. I try to implement this without all the hassle of the additional DST Plugin, which is another PR then.

MilosKozak:
copy from profile switches is now in dev

PR #1549, opened 2018-11-06
Add initial downgrade possibility to this version already

triplem:
In the future we can now increment the DB Version and the user is still able to downgrade.


MilosKozak:
we replace the database. closing for now

PR #1579, opened 2018-12-07
(WIP) Update checker

triplem:
This is a version checker and checks the currently installed version against a (newly created and on all updates to be adopted) version.json.


triplem:
Close this one, since there is an upstream version checker already in place now.

PR #1591, opened 2019-01-02
fix for pump battery age defined in NS

PoweRGbg:


MilosKozak:
pbage is correct

2flea:
Seems there is a little mixup of plugins for pump battery age, NS added BAGE some time ago (https://github.com/nightscout/cgm-remote-monitor/pull/3459) which is working fine for INFO, WARN and URGENT parameters in NS. 
image
image

Careportal and NS client seem to expect some other parameter, PBAGE?!

PR #1593, opened 2019-01-04
Wear bug with Insight-pump: ignore rounding deviations

AdrianLxM:


wootmasterslick:
could you explain a bit what this does?

quizzmaster:
This provides some numerical issues. It's considered to compare the absolute of a difference to a threshold instead of just comparing a difference to zero.

PR #1601, opened 2019-01-15
NS to local Profile copy

PoweRGbg:


triplem:
I did already implemented something like this in #1541. We should join forces on this one. What do you think?

PoweRGbg:
I agree. This could be the step forward in all-in-one profile

MilosKozak:
@triplem 's code seems to be cleaner. could you add missing parts (and add tests if possible)

PoweRGbg:
Closing until I fix the conflicts

PR #1645, opened 2019-02-28
Bugfix: only allow SMB if parent preference is enabled.

jotomo:
Some users reported SMBs being enabled after activating Objective 8, without explicitly enabling SMB.

The pref _Enable SMB_ (key_use_smb) is subject to constraints and is handed over to OpenAPS as _microBolusAllowed_, which is a global flag to enable/disable SMBs. This alone should prevent SMBs if the pref is not set, however this seems to have occurred. I traced the code path but was unable to determine the cause. It should work correctly. There is a test which confirms this.

OpenAPS doses SMBs if the above mentioned flag is true and if either of the more detailed _enableSMB*_ flags is set. This change aligns the input to OpenAPS to what the user sees in the prefs: if the main _Enable SMB_ flag is off, all detailed flags will be off as well. This should effectively prevent the reported behaviour, since OpenAPS only doses SMBs if both _microBolusAllowed_ is true and one of the _enableSMB*_ flags is set.

Manually tested to assert SMBs are dosed if enabled but not when _Enable SMB_ is off, regardless of state of detailed prefs.


jotomo:
Test failures are old failures in _GlucoseStats_. No idea what's up with the new failure in _DanaRS_Packet_APS_History_EventsTest_.

MilosKozak:
that's my fault. I need to find out how to mock it properly

MilosKozak:
I agree it should work. But we add this patch as additional safety
Thanks @jotomo

Issue #1661, opened 2019-03-04
LocalInsight Driver swallows extended bolus requests


fra-orolo:
When using the new Insight driver, the Extended Bolus button on the AKT tab is misleading:
You can press it and enter insulin and time, but it will not do anything.
The implemented method in the driver is only cancelling previous extended boluses, but returning a successful Pump-Result -- thus no negative message for the user.

I know, that extended boluses and looping are somehow controversial and difficult, I also know the Insight driver is using extended boluses for high temp basals > 250% , which might cause conflicts, if a true extended basal is running and a high TBR is scheduled by the loop at the same time.
(Although the pump can deliver at least 2 extended boluses at the same time in parallel.)

But for safety: 

- Either the button tries to deliver the extended bolus (if config or constraints permit),
- or the button should not be there, be disabled 
- or at least return an error / message feedback that nothing will happen in the pump.

In my case I fully implemented the method by forwarding to the code that is used by the simulated high temp basals. 
But the high temp basals are switched off in the config for me. So there is hopefully no conflict with the running loop.

TebbeUbben:
That is indeed a bug and will be fixed.

PR #1662, opened 2019-03-04
Allow extended boluses with button, if not used for TBR

fra-orolo:
This is my attempt to fix

[Issue 1661](https://github.com/MilosKozak/AndroidAPS/issues/1661)


TebbeUbben:
The solution is much simpler, see #1663.

fra-orolo:
Sure, convinced. I was not sure whether this gets really never called in the wrong config.

Issue #1626, opened 2019-02-12
[Enhancement] Automatically run loop when TT is changed


harispoljo:
When I set a new TT the loop will run when next bg comes, therefore worstcase you have to wait 5min for a new tbr. I currently do it manually by going to the loop tab and pressing "run loop" but I figured it can be a automatized. 

I thought about adding this bit of code in the method that changes the TT but couldn't find a good way to implement it. 
```
new Thread(() -> LoopPlugin.getPlugin().invoke("TT changed", true)).start();
        FabricPrivacy.getInstance().logCustom(new CustomEvent("TT_Loop_Run"));
```

wootmasterslick:
This isnt in yet? Deninitely an improvement and easy win.

harispoljo:
I changed it to tt_loop_run

PR #1669, opened 2019-03-06
[Enhancement] Run loop when TemporyTarget is changed

harispoljo:
See #1626

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
@@ -34,6 +34,7 @@ import info.nightscout.androidaps.db.DatabaseHelper;
 import info.nightscout.androidaps.db.Source;
 import info.nightscout.androidaps.db.TemporaryBasal;
 import info.nightscout.androidaps.events.EventNewBG;
+import info.nightscout.androidaps.events.EventTempTargetChange;
 import info.nightscout.androidaps.interfaces.APSInterface;
 import info.nightscout.androidaps.interfaces.Constraint;
 import info.nightscout.androidaps.interfaces.PluginBase;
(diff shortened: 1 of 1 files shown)


MilosKozak:
@AdrianLxM @jotomo 
I personaly have no problem with it. I'm only not sure if there is some significant impact

TebbeUbben:
I'm okay with it as well.

jotomo:
* With a high target, the loop might set a zero temp faster (ok)
* With a low target, within 3 minutes after the last invocation with an SMB, the SMB will be blocked, but a temp basal (possibly zero temp) will be set. This might be confusing as the user expects additional insulin (tolerable). This could be remedied by enforcing the 3 minute timewindow using constraints, so the user sees a message (currently _invoke_ only logs this). 
* With a low target, 3-5 minutes after the last invocation a second SMB (for the same BG) might be administered. This would then block an SMB for the next (BG-invoked) loop iteration (since the previous SMB is within the 3 minute window) (tolerable)
* Concurrent invocations are already possibly with the Loop tab, with the regular safeguards preventing harm (ok)

Overall, the benefit would be ~2.5m faster reaction and an immediately visible reaction to the users request.

app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
```diff
@@ -175,6 +176,14 @@ public long suspendedTo() {
         return loopSuspendedTill;
     }
 
+    @Subscribe
+    public void onStatusEvent(final EventTempTargetChange ev) {
+        new Thread(() -> LoopPlugin.getPlugin().invoke("EventTempTargetChange", true)).start();
```
jotomo: Minor style point: _LoopPlugin.getPlugin()_ is redundant. Calling _invoke_ directly would make the intent easier to read.

PR #1670, opened 2019-03-06
Storage constraint

PoweRGbg:

PR #1673, opened 2019-03-07
storageConstraint

PoweRGbg:
the same as #1670 but no trash commits


MilosKozak:
👍
did you test it in reality?

PoweRGbg:
It's tested in emulator only but device testing will be done after finishing the method tests

PR #1680, opened 2019-03-11
Fix1660

fra-orolo:
fixed reading of BG-data that originates in the future, if phones are not perfectly time-synched. Retrigger drawing of UI, after BG-data has been fetched from DB.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
+++ b/app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
@@ -398,6 +398,7 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
     }
 
     private static void scheduleBgChange(@Nullable final BgReading bgReading) {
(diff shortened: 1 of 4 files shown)


MilosKozak:
so the main issue is wrong date sync between devices?

fra-orolo:
- The missing time sync (by > +2-10 seconds into the future) generates one 5-minutes interval
- the second is the collection of BG-values in a racing thread, while the list of collected values is used already in the overview. 

Thus the worst effect would be values from ~10mins ago on the display.

I believe the computations had only a lag of 5min max. I wonder if we should display some warning on the overview (not only in the logs), but it was too late for this yesterday. 
The time on the phones was auto-synced to the telco network, the difference actually varied. I would be able to sync on NTP (with rooted phone) but how many average users can/would do that?

MilosKozak:
i created little bit cleaner solution
https://github.com/MilosKozak/AndroidAPS/pull/1686
can someone test it?

PR #1685, opened 2019-03-15
Calculating bgReading direction if null

PoweRGbg:


MilosKozak:
this has to be changed little bit. the query must not be only on last BG. we have to find previous record, preferably 5 min back (there are BG sources not having 5 min interval).
Or use bucketed data (alligned to 5 min)

PR #1686, opened 2019-03-17
Fix1660

MilosKozak:


AlexD97:
On my watch it is similar to @zehnBE . So the penultimate value is displayed.

fra-orolo:
The idea of an allowed future-range for the end-timestamp looks reasonable. 
The race-condition is still open. Updates based on recent BG-values should only be done, after the 
the IOBCalculatorPlugin.loadBgData() is really finished. 
Otherwise the displayed values are eventually 5 mins old. (Depending on the load-level and compute power of your phone)

fra-orolo:
I did not change the reading frame originally, because I saw a couple of usages of System.currenttimemillis() in other places of the code, so the "future" measurements might still confuse other parts.
The most correct solution (but complicated) would be some tracking of system time-shift against a server-coordinated time (or NTP).

MilosKozak:
acting on newBG is definitely wrong because most of the code needs finished calculations

MilosKozak:
@fra-orolo if you see possible problem let me know.
and thanks for discovering the issue, only your solution is little bit a hack. This seems to be cleaner

fra-orolo:
@MilosKozak Fix looks definitely better than mine. I'm still learning how the events and cascades are designed in AAPS.

PR #1692, opened 2019-03-28
only run DST helper on certain pumps

AdrianLxM:


TebbeUbben:
Shouldn't we rather add this field to the pump description?

AdrianLxM:
Pump descriptions don't distinguish if it is a real pump or a virtual pump.
It is rather the pump driver that can or can't handle it.

app/src/main/java/info/nightscout/androidaps/plugins/constraints/dstHelper/DstHelperPlugin.java
```diff
@@ -43,83 +42,53 @@ public DstHelperPlugin() {
...
         }
+
+        if (!value.value()) {
+            log.debug("Already not allowed - don't check further");
+            return value;
+        }
+
+        if (wasDST(cal)) {
+            LoopPlugin loopPlugin = LoopPlugin.getPlugin();
+            if (loopPlugin.isSuspended()) {
```
schmitzn: I think it should be:
```suggestion
            if (!loopPlugin.isSuspended()) {
```

MilosKozak:
👍

PR #1701, opened 2019-03-30
Automation

MilosKozak:


Heiner1:
Is there a chance to avouid submitting every profile change, when they are calculated but staying at the same level then before? It causes unreadable reports and may be data garbage.

app/src/main/java/info/nightscout/androidaps/services/LocationService.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    public void onDestroy() {
+        if (L.isEnabled(L.LOCATION))
+            log.debug("onDestroy");
+        super.onDestroy();
+        MainApp.bus().unregister(this);
+        if (mLocationManager != null) {
+            try {
+                if (ActivityCompat.checkSelfPermission(this, android.Manifest.permission.ACCESS_FINE_LOCATION) != PackageManager.PERMISSION_GRANTED && ActivityCompat.checkSelfPermission(this, android.Manifest.permission.ACCESS_COARSE_LOCATION) != PackageManager.PERMISSION_GRANTED) {
```
adrian-hidothealth: Shouldn't this rather be asked when we register for location updates and notify?
Or can this permission implicitly be revoked by toggling location on the phone?

AdrianLxM: Shoudn't we check that when requesting location updates?
Or can they be implicitly revoked by toggling off location on the phone? If yes, shouldn't we still check earlier?

app/src/main/java/info/nightscout/androidaps/services/LocationService.java
```diff
@@ -0,0 +1,146 @@
...
+                mLocationManager.requestLocationUpdates(
+                        LocationManager.NETWORK_PROVIDER,
+                        T.mins(5).msecs(),
+                        LOCATION_DISTANCE,
+                        mLocationListener = new LocationListener(LocationManager.NETWORK_PROVIDER)
+                );
+            if (SP.getString(R.string.key_location, "NONE").equals("GPS"))
+                mLocationManager.requestLocationUpdates(
+                        LocationManager.GPS_PROVIDER,
+                        T.mins(5).msecs(),
```
AdrianLxM: I'd extract them as constants.

app/src/main/java/info/nightscout/androidaps/plugins/general/automation/AutomationPlugin.java
```diff
@@ -0,0 +1,217 @@
...
+                .shortName(R.string.automation_short)
+                .preferencesId(R.xml.pref_automation)
+                .description(R.string.automation_description)
+        );
+    }
+
+    @Override
+    protected void onStart() {
+        Context context = MainApp.instance().getApplicationContext();
+        context.startService(new Intent(context, LocationService.class));
```
AdrianLxM: I don't have the full overview over the automation system yet. But wouldn't it be good to only start it if some component needs those updates?

app/src/main/java/info/nightscout/androidaps/services/LocationService.java
MilosKozak: I think ask when you enable Automation is ok. At this time location service is started

app/src/main/java/info/nightscout/androidaps/plugins/general/automation/AutomationPlugin.java
MilosKozak: done

app/src/main/java/info/nightscout/androidaps/services/LocationService.java
MilosKozak: done

app/src/main/java/info/nightscout/androidaps/plugins/general/automation/AutomationPlugin.java
MilosKozak: I'd not make it too complicated. In passive mode it doesn't cost more energy

Issue #1698, opened 2019-03-29
GlucoseStatus calculates incorrect deltas for 1 minute spaced data


rob-andy:
When using glimp with a miaomiao sensor to feed data into nightscout, the deltas displayed on the overview (average,short&long) are not correct.

This can be traced to the getGlucoseStatusData method of the GlucoseStatus class.

By modifying the unit test to use data such as 
```java
    List<BgReading> generateLibreTestData() {
        List<BgReading> list = new ArrayList<>();
        try {
            double start_time = 1514766900000d;

            list.add(new BgReading(new NSSgv(new JSONObject("{\"mgdl\":100,\"mills\":" + start_time + ",\"direction\":\"Flat\"}"))));

            start_time -= 1000d * 60; // 1 minute ago
            list.add(new BgReading(new NSSgv(new JSONObject("{\"mgdl\":100,\"mills\":" + start_time + ",\"direction\":\"Flat\"}"))));
            start_time -= 1000d * 60; // 2 minutes ago
            list.add(new BgReading(new NSSgv(new JSONObject("{\"mgdl\":100,\"mills\":" + start_time + ",\"direction\":\"Flat\"}"))));
            start_time -= 1000d * 60; // 3 minutes ago
            list.add(new BgReading(new NSSgv(new JSONObject("{\"mgdl\":110,\"mills\":" + start_time + ",\"direction\":\"Flat\"}"))));
            start_time -= 1000d * 60; // 4 minutes ago
            list.add(new BgReading(new NSSgv(new JSONObject("{\"mgdl\":110,\"mills\":" + start_time + ",\"direction\":\"Flat\"}"))));
            start_time -= 1000d * 60; // 5 minutes ago
            list.add(new BgReading(new NSSgv(new JSONObject("{\"mgdl\":110,\"mills\":" + start_time + ",\"direction\":\"Flat\"}"))));
            start_time -= 1000d * 60; // 6 minutes ago
            list.add(new BgReading(new NSSgv(new JSONObject("{\"mgdl\":110,\"mills\":" + start_time + ",\"direction\":\"Flat\"}"))));
            start_time -= 1000d * 60; // 7 minutes ago
            list.add(new BgReading(new NSSgv(new JSONObject("{\"mgdl\":110,\"mills\":" + start_time + ",\"direction\":\"Flat\"}"))));

        } catch (JSONException e) {
            e.printStackTrace();
        }
        return list;
    }
```

[...]

PR #1703, opened 2019-03-31
Calculate deltas correctly with 1 minute spaced data

rob-andy:
* Remove hard limit of processing 9 records.  Processing will now stop
    at first value outside of time range of interest
  * Keep a list of all values within 2.5 minutes.  Average this value to use
    as the now value
  * Don't change the now value date.  Always use the time of the latest value
  * Added new test for data spaced at 1 minute intervals, to mimic libre data from glimp
  * Fix existing test results for average < 2.5 minute data. Deltas are different due to
    not updating the now value time


rob-andy:
For issue #1698

PR #1704, opened 2019-03-31
Insight fix

TebbeUbben:
This fixes the following issues:

- Proper rounding for basal values (the pump returns an error other wise)
- Update pump description to properly match the Insight's limits
- Set bolusDelivered field in PumpEnactResult when delivering bolus
- Fix updating pump time when setting the clock ahead due to DST

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
@@ -421,7 +421,7 @@ public class LocalInsightPlugin extends PluginBase implements PumpInterface, Con
             if (profile.getBasalValues().length > i + 1)
                 nextValue = profile.getBasalValues()[i + 1];
             BasalProfileBlock profileBlock = new BasalProfileBlock();
-            profileBlock.setBasalAmount(basalValue.value);
+            profileBlock.setBasalAmount(basalValue.value > 5 ? Math.round(basalValue.value / 0.1) * 0.1 : Math.round(basalValue.value / 0.01) * 0.01);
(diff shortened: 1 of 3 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -476,7 +476,7 @@ public boolean isThisProfileSet(Profile profile) {
                 nextValue = profile.getBasalValues()[i + 1];
             if (profileBlock.getDuration() * 60 != (nextValue != null ? nextValue.timeAsSeconds : 24 * 60 * 60) - basalValue.timeAsSeconds)
                 return false;
-            if (Math.abs(profileBlock.getBasalAmount() - basalValue.value) > 0.01D)
+            if (Math.abs(profileBlock.getBasalAmount() - basalValue.value) > (basalValue.value > 5 ? 0.1 : 0.01))
```
AdrianLxM: Just checking: is this >5 or >=5?

(profileBlock.getBasalAmount() could be 5.03 and basalValue.value could be 5?)

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -512,6 +512,7 @@ public PumpEnactResult deliverTreatment(DetailedBolusInfo detailedBolusInfo) {
         PumpEnactResult result = new PumpEnactResult();
         if (detailedBolusInfo.insulin > 0) {
             try {
+                detailedBolusInfo.insulin = Math.round(detailedBolusInfo.insulin / 0.01) * 0.01;
```
AdrianLxM: Not sure if it is good to change a parameter passed to a method that the caller is not expecting to  be changed?
(should not matter in this case)

AdrianLxM:
lgtm

PR #1725, opened 2019-04-08
fix error with big sms text in UTF-8

Lexsus:


Lexsus:
in Russian language when i send BG command, i not reseived answer.

MilosKozak:
use proper formating - press Ctrl - Alt - L in studio

Lexsus:
applied formatting

Issue #1740, opened 2019-04-12
Local Broadcast to xDrip+ broken?


swissalpine:
After installing the actual dev build this morning I lost the connection to xDrip+, no sgv are transmitted to AAPS. I controlled the settings but nothing has changed. Reinstalling an older version reenables the communication.

st0ber:
After I installed the last Dev version, I did not get any blood glucose levels from the patched Dexcom G5 (not the 2018 version). Is this possibly the same problem? Afterwards I installed the master version again and it worked again.

PR #1742, opened 2019-04-13
Split up receivers for local BG broadcasts and SMS.

swissalpine:
Receiving local BG values probably fails silently when SMS
permission is not granted (due to SMS plugin not being used).

Many thanks (once more) to Jotomo who has fixed https://github.com/MilosKozak/AndroidAPS/issues/1740

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -73,31 +73,40 @@
         <activity android:name=".plugins.pump.danaRS.activities.PairingHelperActivity" />
         <activity android:name=".activities.HistoryBrowseActivity" />
 
+        <!-- Receive new BG readings from other local aaps -->
         <receiver
-            android:name=".receivers.DataReceiver"
-            android:enabled="true"
-            android:exported="true"
-            android:permission="android.permission.BROADCAST_SMS">
+                android:name=".receivers.DataReceiver"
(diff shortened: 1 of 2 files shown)


AdrianLxM:
LGTM. Could you please add a comment to SMSReceiver why you have it as a separate class.
And please add a TODO that the functionality actually gets refactored into the correct class.

Thanks

MilosKozak:
is it a new feature of android? it works current way for 3 years now

jotomo:
@MilosKozak The addition of the BROADCAST_SMS permission to the _DataReceiver (https://github.com/MilosKozak/AndroidAPS/commit/e95e7922b306b5f91241ec690d3dd843d5ff5327#diff-c8cc2dd42271f2bf39c6aa81eb6a5529) meant that for users not using the SMS plugin (and thus not having granted the app the SMS permission), the receiver wouldn't work at all, thus xDrip etc. broadcasts weren't received anymore as per #1740, #1744 .
This PR creates a separate SMS receiver in the manifest and accompanying receiver class.
@AdrianLxM Yup, comment should be there :-)
The new class also simply forwards to _DataService_, which is basically a dispatcher to forward messages to the enabled (BG Source) plugins. What functionality would you want to have moved (and where)?

MilosKozak:
@jotomo thanks!

PR #1757, opened 2019-04-22
update logic version checker

AdrianLxM:


app/src/main/java/info/nightscout/androidaps/plugins/general/versionChecker/VersionCheckerUtils.kt
```diff
@@ -56,23 +65,31 @@ fun compareWithCurrentVersion(newVersion: String?, currentVersion: String) {
...
 }
 
 fun onNewVersionDetected(currentVersion: String, newVersion: String?) {
-    log.debug("Version ${currentVersion} outdated. Found $newVersion")
-    val notification = Notification(Notification.NEWVERSIONDETECTED, String.format(MainApp.gs(R.string.versionavailable), newVersion.toString()), Notification.LOW)
-    MainApp.bus().post(EventNewNotification(notification))
-    SP.putLong(R.string.key_new_version_available_since, System.currentTimeMillis())
+    val now = System.currentTimeMillis()
+    if(now > SP.getLong(R.string.key_last_versionchecker_warning, 0) + WARN_EVERY) {
+        log.debug("Version ${currentVersion} outdated. Found $newVersion")
```
MilosKozak: both notations are ok?

AdrianLxM: That's how AndroidStudio inlined it.
With braces it evaluates the code, without it only takes the content of the variable.
So yea, it is ok like this. Still wondering why AndroidStudio did this.

PR #1758, opened 2019-04-23
try to calculate direction of BG if missing

PoweRGbg:


MilosKozak:
It looks much better. Could you add some tests?

PoweRGbg:
@MilosKozak Ready for merge ?

MilosKozak:
created PR with cleanup https://github.com/PoweRGbg/AndroidAPS/pull/68 . Merge it pls
but it still need update. The call for direction calculation may be performed for older records as well so it's not simply possible use last glucose status

app/src/test/java/info/nightscout/androidaps/db/BgReadingTest.java
```diff
@@ -0,0 +1,168 @@
...
+        assertTrue(copy.isEqual(bgReading));
+        assertFalse(copy.isEqual(new BgReading()));
+    }
+
+    @Test
+    public void calculateDirection() {
+        List<BgReading> bgReadingsList = null;
+        AAPSMocker.mockDatabaseHelper();
+
+        when(MainApp.getDbHelper().getAllBgreadingsDataFromTime(anyLong(),anyBoolean())).thenReturn(bgReadingsList);
```
MilosKozak: this `when` is necesary only once

PoweRGbg: When I remove the others - test fails

PR #1762, opened 2019-04-24
2.3

AdrianLxM:


AdrianLxM:
* Important safety fix for Insight (really important if you use Insight!)
* Fix History-Browser
* Language updates
* Check for GIT and warn on gradle upgrade
* More automatic testing
* Fixing potential crash in AlarmSound Service (thanks @lee-b !)
* Fix broadcast of BG data (works independently of SMS permission now!)
* New Version-Checker

PR #1781, opened 2019-04-29
NPE fixes from Firebase

MilosKozak:


app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/activities/InsightPairingActivity.java
```diff
@@ -160,20 +160,25 @@ public void onStateChanged(InsightState state) {
...
-            registerReceiver(broadcastReceiver, intentFilter);
-            bluetoothAdapter.startDiscovery();
+            if (bluetoothAdapter != null) {
+                if (!bluetoothAdapter.isEnabled()) bluetoothAdapter.enable();
+                IntentFilter intentFilter = new IntentFilter();
+                intentFilter.addAction(BluetoothAdapter.ACTION_DISCOVERY_FINISHED);
+                intentFilter.addAction(BluetoothDevice.ACTION_FOUND);
+                registerReceiver(broadcastReceiver, intentFilter);
+                bluetoothAdapter.startDiscovery();
+            }
```
AdrianLxM: IMHO it should return false in case of null.

app/src/main/java/info/nightscout/androidaps/queue/CommandQueue.java
```diff
@@ -126,6 +128,7 @@ private synchronized void add(Command command) {
         queue.add(command);
     }
 
+    @Nullable
```
AdrianLxM: Why `@Nullable ` on a `void` method?

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/activities/InsightPairingActivity.java
MilosKozak: what exactly?

app/src/main/java/info/nightscout/androidaps/queue/CommandQueue.java
MilosKozak: thanks. bad day

PR #1786, opened 2019-05-01
Signature Verifier

TebbeUbben:


TebbeUbben:
> add tests. the idea looks good

Testing this is actually quite hard because the critical parts require a full Android runtime running an APK signed with a specific certificate to be tested.

MilosKozak:
mock system interaction

PR #1812, opened 2019-05-22
Fmegen/updating color scheme

fmegen:
This pullrequests add an additional level of color indication to the AndroidAPS application as well as the Circle watchface.

Coloring is analog to Nightscout, i.e., in addition to a yellow marker for values that are above the normal range, a new color (red by default) is added for values that are very high (default is 260d).

please note that I did not do translations for the new setting other than english.


app/src/main/res/values/strings.xml
```diff
@@ -418,6 +418,7 @@
     <string name="prefs_range_title">Range for Visualization</string>
     <string name="prefs_range_summary">High and low mark for the charts in Overview and Smartwatch</string>
     <string name="low_mark">LOW mark</string>
+    <string name="veryhigh_mark">VERY HIGH mark</string>
```
fmegen: please note that this needs to be translated to other supported languages

wear/src/main/java/info/nightscout/androidaps/watchfaces/CircleWatchface.java
```diff
@@ -524,7 +534,7 @@ public void onReceive(Context context, Intent intent) {
             Bundle bundle = intent.getBundleExtra("data");
             if (bundle != null) {
                 DataMap dataMap = DataMap.fromBundle(bundle);
-                setSgvLevel((int) dataMap.getLong("sgvLevel"));
+                setSgvLevel((int) dataMap.getLong("sgvLevel2"));
```
fmegen: if all watchfaces are to support the new level, this might be in the normal "sgvLevel"

MilosKozak:
@AdrianLxM are you OK with this?

AdrianLxM:
For Design please refer to @TebbeUbben

wootmasterslick:
Any screenshots of the watchface available?

fmegen:
Here is an example of the watchface.
image
The interesting change is the red marker on the left which indicates the "too high" range (In this case, 260). If the value is higher, the outer ring changes to red too.

Here is an example of the settings page where the additional "too high" setting can be configured.
image

fmegen:
I forgot, here is an example of the main UI,
in the picture, the "very high" value is set to 240

image

fmegen:
i updated the branch to be based on v2.5.0

fmegen:
@MilosKozak I fixed the compilation errors.
kind regards,
 friedel

fmegen:
@MilosKozak  any update?

PR #1829, opened 2019-06-12
Changes to treatment handling

andyrozman:
Changes to treatment handling, replaced createOrUpdate method with little bit more compact method, that
covers all options and has additional method to copy data.


MilosKozak:
This change the logic a lot. I'm really not sure about impact of this change. What is the purpose of this change?

MilosKozak:
@andyrozman this can be closed, right?

andyrozman:
Yes this can be closed. All the changes are in medtronic pull and we also modified those... (2 separate methods one for MDT and one for other...

PR #1854, opened 2019-06-30
DataService: use JobIntentService instead of deprecated WakefulBroadc…

jotomo:
…astReceiver.


jotomo:
JobIntentService is unsuited, as it is designed for deferrable work but incoming SGs need to be processed immediately.

PR #1861, opened 2019-07-05
Show insulin activity in BG graph

Paulusma:
Displays insulin activity (history & prediction) in BG graph as a thin white line.

I think a fixed scale for IA is required to allow a user to compare current IA with that from previous situations. Helps gaining insight in IA trends vs BG trends.

Final optionality & location of IA (could also be included in secondary graph) should be discussed including current performance overhead due to duplicate calls to iobCobCalculatorPlugin.calculateFromTreatmentsAndTempsSynchronized (addActivity is a copy of addIob)

Issue #1851, opened 2019-06-29
Show insulin activity on BG graph


Paulusma:
According to Dana M. Lewis the level of insulin activity is a significant predictor for post-meal BG rise, more so then starting BG level (pge 100 in her book).

If so displaying insulin activity in the main (BG) graph would be of great help in reducing post-meal spikes.

Note: i created such an option in my local version of APS (base V2.3), i've included a screenshot to illustrate the request. The white line is insulin activity, scale is fixed.

![Screenshot_20190629-160340](https://user-images.githubusercontent.com/2260419/60385255-9d30e300-9a87-11e9-89c1-c0f075a3f9d3.png)

Paulusma:
PS: i'm aware of the 'eating soon' option, my request helps the user decide if the 'eating soon' target is needed or not (graph also predicts insulin activity).

fra-orolo:
I would like this plot very much.
@Paulusma If you have it in your local version, could prepare a patch / pull request?

Paulusma:
OK but it will take some time, i've never done that.

Paulusma:
Patch wrt latest master included with 2 changes. Note: in this patch IA is not optional. 
[Display_insulin_activity_in_BG_graph_.zip](https://github.com/MilosKozak/AndroidAPS/files/3352038/Display_insulin_activity_in_BG_graph_.zip)

MilosKozak:
create PR please

fra-orolo:
Good stuff, seems to work well: I just added the menu entry for the new curve and made it switchable.
Need to think about the scaling a bit deeper, the magic 1 / 0.04 produced a too big graph for me.

Paulusma:
Thx for the switch! 

I changed form mmol to mg/dl and noticed no difference. Off-scale can be fixed by making scale dependent on max activity that can safely be achieved, which i think can be caluclated form the max IOB and max bolus U preference settings. This could be the fixed upper limit, but i think a reduction to some %% of this value will be required because during normal operation you will probably be way off this max value. I'll look into that.

Paulusma:
OK insulin activity is in U/min so independent of mg/dl vs mmol setting.
Also determining max activity that can be achieved depends on a lot of additional factors, at least also on type of user, and SMB vs AMA. I think to incorporate that makes the code to determine scale too complex/dependent on other algos. I think it's best to include a new preference setting for the scal of activity in BG graph with a sensible default.

Paulusma:
[...]

PR #1867, opened 2019-07-11
Insulin Activity in bg graph

Paulusma:
(Cancelled previous pull request #1861, commits got messy)

#1851 Show insulin activity in BG graph as a thin solid (history) / dashed (prediction) white line.

- Configurable through BG overview popup menu.
- fixed scale for insulin activity is configurable through preference 'key_scale_insulin_activity' (default 0.05)


app/src/main/java/info/nightscout/androidaps/plugins/general/overview/graphData/GraphData.java
```diff
@@ -344,6 +346,55 @@ private double getNearestBg(long date) {
...
+        List<ScaledDataPoint> actArrayPred = new ArrayList<>();
+
+        double now = System.currentTimeMillis();
+        Scale actScale = new Scale();
+        IobTotal total = null;
+
+        for (long time = fromTime; time <= toTime; time += 5 * 60 * 1000L) {
+            Profile profile = ProfileFunctions.getInstance().getProfile(time);
+            double act = 0d;
+            if (profile != null)
```
jotomo: This only defers the NPE (till _total.activity_ a few lines down). The profile being null is an edge-case that occurs when the app is freshly setup and the graph shows a timeframe in the past where no profile (switch) was active. So the best thing to do here is skip the iteration (``` if (profile == null) continue; ```)

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/graphData/GraphData.java
```diff
@@ -344,6 +346,55 @@ private double getNearestBg(long date) {
...
+                total = iobCobCalculatorPlugin.calculateFromTreatmentsAndTempsSynchronized(time, profile);
```
jotomo: For data in the future this may introduce a significant performance regression since calculations in the future aren't cached (see the unsynchronized method in the IobCob... class). This should be noticeable when changing the displayed timeframe (longpress on the graph) with and without IA shown.

jotomo:
@Paulusma Liko IOB, IA can become negative. The GraphView widget will likely simply not show negative data as only a positive range is displayed. However, in the second chart, IA might fit better (as a switchable alternative to IOB). There, it could show how insulin action and carb deviations match up, e.g. if the timing is good, since that's what IA shows over IOB.

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/graphData/GraphData.java
Paulusma: @jotomo thx updated the code accordingly.

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/graphData/GraphData.java
Paulusma: @jotomo @MilosKozak I'm not sure which of these options is best here: 

1) Including IA in BG graph makes it easier to compare IA <-> BG *but* there is the additional overhead, and IA can become negative. 

2) Including IA in IOB graph has almost no performance penalty if we calculate both IOB and IA in  the some loop (ofcourse only when selected) and allows better comparison of IOB, deviations etc  with IA.

To me comparison of BG <-> IA seems of most interest,  so i would prefer 1).

What are your thoughts?

app/src/main/res/values-af/strings.xml
```diff
@@ -1149,6 +1149,8 @@
     <string name="sms_wrongcode">Verkeerde kode. Opdrag gekanselleer.</string>
     <string name="notconfigured">Nie gekonfigureer nie</string>
     <string name="profileswitchcreated">Profiel skakelaar geskep</string>
+    <string name="key_scale_insulin_activity">key_scale_insulin_activity</string>
```
MilosKozak: <string name="key_scale_insulin_activity" formatted="false">scale_insulin_activity</string>

app/src/main/res/values-af/strings.xml
```diff
@@ -1149,6 +1149,8 @@
...
+    <string name="scale_insulin_activity">Graph scale insulin activity [U/min]</string>
```
MilosKozak: + remove other languages of both lines ...

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/graphData/GraphData.java
MilosKozak: i tested it now and there is no visible impact on performance

MilosKozak:
the same code shuld be added to History browser

MilosKozak:
negative activity is not displayed but i don't see it as a big issue. negative graph will be always pretty flat

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/graphData/GraphData.java
jotomo: @Paulusma
1) While charting in the top graph more directly shows impact on BG, it _only_ shows effect of IA on BG. What I found helpful in working out meal timing was overlaying IA with deviations and COB to see how they balance out. E.g. in this experimental visualization you can IMO nicely see the effect of Eating Soon and how IA and deviations (black) match up and battle it out to maintain a stable BG (cherry-picked example, only works during a full moon ;-) ):
ia-vs-devs

2) The cost of accessing the data in multiple loops is negligible, the cost comes from insulin maths not being cached for future dates. I haven't tried how much difference it makes with this PR, but have seen slowdowns where I experimented with this ... okay, I've just tested it, it doesn't make a noticeable difference (on a SD820 SoC). Negative IA is simply cut off:
Screenshot_20190713-153702_AndroidAPS

For situations where only insulin is active (e.g. during the night, mostly), showing IA in the top graph could (also) be nice to see IA-only impact and tweak insulin duration setting 
... do both, use it and see which one works out for you? That's often my approach :-)

jotomo: @MilosKozak Thanks for testing, I took to long to write the post :-)

jotomo:
> negative activity is not displayed but i don't see it as a big issue. negative graph will be always pretty flat

Correct. I wanted to make sure it doesn't cause any funny errors.

app/src/main/res/values-af/strings.xml
Paulusma: done

app/src/main/res/values-af/strings.xml
Paulusma: done

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/graphData/GraphData.java
Paulusma: I included an option to display IA in secondary graph as well. User can decide on wich graph to display IA (or even on both). 

@jotomo @MilosKozak could you please comment on this?

Paulusma:
> the same code should be added to History browser
done

Paulusma:
> the same code should be added to History browser

done!

Paulusma:
> the same code should be added to History browser

done!

Paulusma:
> > the same code should be added to History browser
> 
> done!

gdfgg

Paulusma:
> the same code should be added to History browser

Done, but i notice that IA is zero for historic data, much like IOB???

MilosKozak:
history browser will be updated after db change. currently it's not perfect

Paulusma:
@MilosKozak dev has a *double* divider for the BG graph popup menu. Historybrowser is OK.

PR #1868, opened 2019-07-12
Medtronic driver merge

andyrozman:
This is implementation of Medtronic driver. For now its hidden under engineering mode. Documentation will be merged soon.


jotomo:
Some notes from a very quick first look:
TreatmentsPlugin: why all the style changes? Makes it hard to see what has actually changed.
README: still caries the MD-branch specific instructions
FabricPrivacy: still has debug code
logback: revert debug changes?

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/dialogs/MessageDialog.java
```diff
@@ -0,0 +1,90 @@
+package info.nightscout.androidaps.plugins.general.overview.dialogs;
```
MilosKozak: Isn't  it duplicate to `OKDialog`?

app/src/main/java/info/nightscout/androidaps/MainApp.java
```diff
@@ -252,6 +264,54 @@ private void registerLocalBroadcastReceiver() {
         lbm.registerReceiver(dbAccessReciever, new IntentFilter(Intents.ACTION_DATABASE));
     }
 
+
+    private void setBTReceiver() {
```
MilosKozak: Can this be standart receiver defined in manifest?

app/src/main/java/info/nightscout/androidaps/MainApp.java
```diff
@@ -252,6 +264,54 @@ private void registerLocalBroadcastReceiver() {
...
+                            case BluetoothAdapter.STATE_ON: {
+                                if ("Medtronic".equals(activePump.deviceID())) {
+                                    Log.v("MainApp", "Bluetooth on");
+                                    RileyLinkUtil.sendBroadcastMessage(RileyLinkConst.Intents.BluetoothReconnected);
+                                }
+                            }
+                            break;
+                        }
+                    } else {
+                        activePump.timeDateOrTimeZoneChanged();
```
MilosKozak: How is this related to timezone change?

app/src/main/java/info/nightscout/androidaps/plugins/pump/common/defs/PumpType.java.orig
```diff
@@ -0,0 +1,398 @@
+package info.nightscout.androidaps.plugins.pump.common.defs;
```
MilosKozak: redundant?

app/src/test/java/info/nightscout/androidaps/utils/LoggerRule.java
```diff
@@ -0,0 +1,62 @@
+package info.nightscout.androidaps.utils;
```
MilosKozak: ???

app/src/test/java/android/util/Log.java
```diff
@@ -0,0 +1,59 @@
+package android.util;
```
MilosKozak: Can this be transformed to use Logger as everywhere else?

app/src/main/res/xml/pref_medtronic.xml
```diff
@@ -0,0 +1,81 @@
+<?xml version="1.0" encoding="utf-8"?>
+<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
+    <PreferenceCategory android:title="@string/medtronic_name">
+
+
+        <EditTextPreference
+            android:defaultValue="000000"
+            android:key="pref_medtronic_serial"
```
MilosKozak: use  `key_...` like everywhere alse to avoid typos

app/src/main/res/layout/overview_wizard_dialog_compact.xml
```diff
@@ -0,0 +1,522 @@
+<?xml version="1.0" encoding="utf-8"?>
```
MilosKozak: not referenced

app/src/main/java/info/nightscout/androidaps/plugins/treatments/TreatmentsPlugin.java
```diff
@@ -550,6 +579,10 @@ public boolean addToHistoryTreatment(DetailedBolusInfo detailedBolusInfo, boolea
         return newRecordCreated;
     }
 
+    private boolean isMedtronicPump() {
```
MilosKozak: better use `MedtronicPlugin.getInstance().isEnabled(PluginType.PUMP)`

app/src/main/java/info/nightscout/androidaps/plugins/pump/virtual/VirtualPumpStatus.java
```diff
@@ -0,0 +1,33 @@
+package info.nightscout.androidaps.plugins.pump.virtual;
```
MilosKozak: not used

app/src/main/java/info/nightscout/androidaps/MainApp.java
andyrozman: Maybe. I have no deep understanding how Broadcast receivers work...

app/src/main/java/info/nightscout/androidaps/MainApp.java
andyrozman: I register 4 intents, one is for BT other three are for TZ/Time/Date changes... I wanted to do writing to Db too, but for now this was enough...

app/src/main/java/info/nightscout/androidaps/plugins/pump/common/defs/PumpType.java.orig
andyrozman: what is redundant?

andyrozman: I see what you mean. Will fix.

app/src/test/java/android/util/Log.java
andyrozman: This was for Unit Tests.. I am not able to see any Logger results if I run tests... If I override Log java in test, you can at least follow this log entries...

app/src/main/java/info/nightscout/androidaps/plugins/pump/virtual/VirtualPumpStatus.java
andyrozman: Will remove...

app/src/main/res/layout/overview_wizard_dialog_compact.xml
andyrozman: Will remove

app/src/test/java/info/nightscout/androidaps/utils/LoggerRule.java
andyrozman: Will remove... One of several tries to get Logger to start printing logs when running tests...

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/dialogs/MessageDialog.java
andyrozman: I will try to use OKDialog, and if it works same, we can remove this...

andyrozman: OKDialog doesn't work... It is just not displayed... I modified MessageHelperActivity to call OKDialog and it just didn't work...

app/src/main/java/info/nightscout/androidaps/plugins/treatments/TreatmentsPlugin.java
andyrozman: Would that work always?

app/src/main/res/xml/pref_medtronic.xml
andyrozman: Insight doesn't do it...

app/src/main/java/info/nightscout/androidaps/plugins/treatments/TreatmentsPlugin.java
andyrozman: it seems it does...

app/src/main/java/info/nightscout/androidaps/plugins/pump/common/PumpPluginAbstract.java
```diff
@@ -0,0 +1,446 @@
...
+import info.nightscout.androidaps.utils.DateUtil;
+import info.nightscout.androidaps.utils.DecimalFormatter;
+
+/**
+ * Created by andy on 23.04.18.
+ */
+
+// When using this class, make sure that your first step is to create mConnection (see MedtronicPumpPlugin)
+
+public abstract class PumpPluginAbstract extends PluginBase implements PumpInterface, ConstraintsInterface {
```
jotomo: This class is a MDT-specific base class, it should not be in the _common_ package.

app/src/main/java/info/nightscout/androidaps/MainApp.java
jotomo: This mixes too many things. Calling it a BTReceiver when it does a whole lot more is confusing :-)
While we don't have clean plugin/pump lifecycles yet (Combo is very guilty of that, binding to ruffy service on app start), we should aim for it and separate pump plugins. 
This receiver will trigger Riley code no matter which pump driver is active (and couple MainApp to a RileyLink impl class). I would suggest moving the BT-receiver to MDT-land and register it dynamically (not in the manifest, since it's not always needed) in _onStart_ and unregister in _onStop_.
Handling timezone / date changes _might_ be something to be done on a level above pump drivers and just have the drivers provide a _setDateTime_ method, but that probably won't work since drivers handle time changes differently (this should be discussed, but not as part of a big merge like this). So keeping a receiver here (registered in manifest) to receive time changes and informing the drivers seems a good place to start.

app/src/main/java/info/nightscout/androidaps/utils/SP.java
```diff
@@ -132,6 +132,19 @@ static public void putInt(int resourceID, int value) {
...
+        editor.putString(MainApp.gs(resourceID), value.toString());
+        editor.apply();
+    }
+
+    static public void putDouble(String key, Double value) {
+        SharedPreferences.Editor editor = sharedPreferences.edit();
+        editor.putString(key, value.toString());
+        editor.apply();
+    }
+
```
jotomo: Can be removed, they're not used (they're overloading version with a primitive double as value) and will fail for null-values.

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/dialogs/MessageDialog.java
jotomo: Locking at invocation sites ... is the helper activity needed? Callers already open a new thread, so there's no blocking issue and _OkDialog.show()_ (not sure if this blocks) could be invoked directly.
However, the invocation in _MedtronicUtil.displayNotConfiguredDialog_:  messages like these are usually displayed as notifications on overview (e.g. _ComboPlugin.checkPumpTime_)
The invocation at _MedtronicPumpPlugin.deliverBolus()_: shouldn't this integrate with _BolusProgressDialog_ since it happens during bolusing? (Don't put time into issues you might see with that dialog, esp. with SDK upgrades this dialog has nowhere to go but being replaced with a progress notification).
With that, neither helper nor _MessageDialog_ itself would be needed anymore if I'm not overlooking something.

andyrozman: displayNotConfiguredDialog - This is for dialogs and actions. We have several dialogs and actions, and calling them when device was not configured produced Exceptions... Putting warning into Overview would be useless here, because User is not in Overview, so he wouldn't see warning until next time. I was testing situation with Bolus, where we also call this dialog, and OKDialog doesn't work there... 
deliverBolus - To get full picture you need to look at method deliverTreatment in PumpPluginAbstract... that one actually calls deliverBolus...

app/src/main/java/info/nightscout/androidaps/MainApp.java
andyrozman: I tried to register it in the pump itself and it just didn't work there... None of BT events was received... Maybe I did mistake somewhere or forgot to register something, but all BT events were ignored when run from Driver itself, I tried several other options (I was playing around with this for few days). It started working when I registered it in MainApp, similar as you do all other Receivers. I could have made some mistake somewhere, I am relatively new to Android programming. 

As for time changes... That is intended for *any* pump class. I extended PumpInterface with timeDateOrTimeZoneChanged() method. It has empty implementation for all drivers at the moment, but this is the point where driver developers can do something when time changes.. We could do separate receiver here or something, I am just big fan or reusing stuff, instead of creating new instances of same stuff that waste memory...

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/dialogs/MessageDialog.java
andyrozman: That method deliverTreatment was taken from Combo implementation and refactored so that deliverBolus is called (in actual pump driver) when real bolus delivering is going on, and other options in other cases (when no real bolus will be triggered).

jotomo: https://github.com/andyrozman/AndroidAPS/pull/5

app/src/main/java/info/nightscout/androidaps/MainApp.java
jotomo: Re: time changes: definitely reuse code. My thought was whether it might make sense to also pull the adjustment logic from drivers to a level above, stuff like gradually adjusting time or deciding the jump is to big and annoy the user. Extracting that logic to be above the drivers and have the drivers implement dumber logic that only sets the time. I only wish I could set the time on the combo programmatically ;-) In short: that change is fine, I was just derailing the conversation a bit with ponderings going a bit further.

Re: broadcast receivers: not sure what they require to work or setup, still learning Android stuff myself. We have the DataReceiver as a sample for receiving broadcast ... If it needs to stay in MainApp I'd prefer the code to send out an event to which RileyLink etc can subscribe to, rather than have a hard dependency to it in MainApp.

app/src/main/java/info/nightscout/androidaps/plugins/pump/common/PumpPluginAbstract.java
MilosKozak: 👍

app/src/main/java/info/nightscout/androidaps/MainApp.java
andyrozman: time changes:
Yeah, I was thinking to store this time/date/TZ changes into database... That would make synchronization of data much easier (if pump would need it)... 

Receivers:
I tried using them in Manifest and that didn't work correctly... Maybe I did something wrong I am not sure... So I reverted the change... I could make BroadcastReceiver for both: BT and Date/TZ changes and then just create them in MainApp... That would make it little cleaner...

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/dialogs/MessageDialog.java
andyrozman: Merged and we can resolve this

app/src/main/res/xml/pref_medtronic.xml
andyrozman: So is this required to do? I don't think we gain much to add key into string, but you are the experts... 
I don't use this keys directly in code anywhere (I use MedtronicConst and RileyLinkConst classes)

app/src/main/java/info/nightscout/androidaps/MainApp.java
andyrozman: No. Doesn't work if defined in Manifest.

app/src/main/java/info/nightscout/androidaps/plugins/pump/common/PumpPluginAbstract.java
andyrozman: Its not MDT specific... its not any type of pump specific... its abstract class for pump interface...

app/src/main/java/info/nightscout/androidaps/utils/SP.java
andyrozman: removed xxxDouble methods not used.

app/src/main/java/info/nightscout/androidaps/MainApp.java
andyrozman: See next comment about refctoring

app/src/main/java/info/nightscout/androidaps/MainApp.java
andyrozman: So I refactored this a little... I created 2 broadcast receievers... One is registered from MainApp (with MainApp) - Time Change, and other is registered from RileyLinkService (on create) with MainApp.

So I think this is fixed...

app/src/main/res/xml/pref_medtronic.xml
MilosKozak: Using keys prevents typo mistakes. It happend many times in AAPS project

app/src/main/java/info/nightscout/androidaps/plugins/pump/common/PumpPluginAbstract.java
MilosKozak: but it's used only for MDT right now. i'd move it for now

app/src/main/java/info/nightscout/androidaps/plugins/pump/common/data/PumpStatus.java
```diff
@@ -0,0 +1,90 @@
+package info.nightscout.androidaps.plugins.pump.common.data;
```
MilosKozak: PumpStatus is currently MDT specific too

app/src/main/res/values/strings.xml
```diff
@@ -1391,6 +1391,16 @@
     <string name="description_pump_medtronic">Pump integration for Medtronic, requires RileyLink device and specific Pump Model</string>
 
     <!-- MDT Configuration -->
+    <string name="pref_key_medtronic_serial">pref_medtronic_serial</string>
```
MilosKozak: ` translatable="false"` is missing

app/src/main/java/info/nightscout/androidaps/plugins/pump/common/PumpPluginAbstract.java
jotomo: From what I've seen it extracts things from MDT, including things that don't apply to all pumps. The _ComboPlugin_ doesn't (directly) use a service connection for example. And TBH, the initialization is a tad too complex for my taste, being stretched out over the constructors of the abstract class and the concrete sublass, fields of both classes, the _onStart_, _onCustomStart_ as well as _initPumpStatusData_ methods (while AAPS triggers a pump init by issuing a READ_STATE command). Not trying to bash your code, the Combo code is also not as clear in that regard as I'd like it to be (a spike to evaluate a refactoring towards a cleaner lifecycle with the Combo is on my agenda).
However, I'm getting a bit carried away as far as the MDT-merge itself is concerned (which I don't want to delay over this issue): I think the lifecycle of pump plugins could do with a rework (to actually only init a pump when a plugin is enabled and not init all pump plugins on app start). For such a refactoring, having a base class that simplifies and helps with implementing a clean lifecycle might well be helpful.

andyrozman: I intended to use this same abstract class for at least Omnipod and OmnipodDash... but you are right we need to rethink how this things are done... 

I will be working on Omnipod integration with Bart soon (after we finish the merge) and I don't want to duplicate the code where its not needed.

And if everything goes right I will finally get my hands on some Dash pods, so maybe that development will also move from deadpoint...

app/src/main/res/values/strings.xml
andyrozman: Will change... give me half an hour...

PR #1879, opened 2019-07-19
ProfileSwitch validation

MilosKozak:


MilosKozak:
Currently it's possible to bypass NumberPicker limits by entering own number
I added validation on loosing focus but this is still not called when you press OK.
Any ideas?
Better would be to have this covered inside class. Othervise we'd have to validate in every dialog.
Validatin in onTextChanged is not good because it's acting already on first digit (for example when you want to type "35" you get alet at "3")

PR #1906, opened 2019-07-28
use newer version of weekdays-selector, fixes #1904

### LINKED ISSUE #1904: [2.4-alpha1] Wrong title during installation
During installation the screen displays the wrong application name: "WeekdaysSelector" instead of AndroidAPS
Screenshot_20190728-185332

quizzmaster:
Thanks to @jotomo who mentinoned this solution.
This will fix the wrong app name #1904


jotomo:
@quizzmaster Did you check the automation plugin still works (which uses the lib) and the name issue is fixed? I only looked into the cause, but didn't really test solution.

PR #1932, opened 2019-08-02
Check we have battery optimisation with PowerManager

twain47:
Required in addition to REQUEST_IGNORE_BATTERY_OPTIMIZATIONS permission.


twain47:
This isn't quite complete - I can't see how to trigger SetupWizardActivity.onRequestPermissionsResult with the result that it doesn't show the 'next' button after permission is granted.

If anyone has any suggestions on how to implement that I'll add it.

PR #1939, opened 2019-08-05
Remove butterknife dependency

MilosKozak:


MilosKozak:
@AdrianLxM 
removed butterknife where used and affected fragments refactored to kotlin
Events used only in these fragments moved to RxBus
only UI code affected

Issue #1946, opened 2019-08-08
[Medtronic] Bolus reports success when reservoir is empty


twain47:
Boluses are reported as successful by the pump when the reservoir is empty.

Possible work around would be to check it with current value of reservoir. This is updated every hour and all boluses issued are subtracted. If there is not enough insulin left we would either prevent the bolus and error, or log a warning that the bolus may not have been delivered.

PR #1956, opened 2019-08-12
- added check before bolus is attempted, if we have enough in reservo…

andyrozman:
Fix for #1946


app/src/main/res/values/strings.xml
```diff
@@ -1637,6 +1637,7 @@
     <string name="medtronic_cmd_cant_cancel_tbr_stop_op">Could not cancel current TBR. Stopping operation.</string>
     <string name="medtronic_cmd_set_profile_pattern_overflow">Profile set failed, because following patterns, have too big basal rate: %1$s</string>
     <string name="medtronic_cmd_bolus_could_not_be_delivered">Bolus could not be delivered.</string>
+    <string name="medtronic_cmd_bolus_could_not_be_delivered_no_insulin">Bolus could not be delivered, because available amount (%1$d) is less than Bolus required (%2$d).</string>
```
MilosKozak: should not this be a float number?

andyrozman: Yes.

PR #1950, opened 2019-08-10
Update oref master

Tornado-Tim:
This PR brings AAPS up to date with Oref Master. (other PRs made prior are closed in favor of this)

Minor tweaks to determine basal were required as `console.error()` was no longer used upstream and instead was replaced by a node js reporting process function.
The tweaks specifically put the console.logs back into place and does not change anything else.

Noise/Raw data is used in oref master now, I have added prelim support in AAPS with this, but it is disabled as not all CGM's do not report noise an will be effectively off in the first place.

If noise is wanted (for example xdrip which broadcasts this) it would be very easy to enable and just link the glucose plugin through, let me know with this is wanted and I can add it in.

Cheers

Tim

PR #1974, opened 2019-08-21
Display carbs required

twain47:
Add option to expose carbs suggestions provided by the SMB algorithm as a local notification.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -113,6 +113,9 @@
         <!-- Receiver keepalive, scheduled every 30 min -->
         <receiver android:name=".receivers.KeepAliveReceiver" />
 
+        <!-- Receive ignore 5m, 15m, 30m requests for carb notifications -->
+        <receiver android:name=".plugins.aps.loop.CarbSuggestionReceiver"></receiver>
+
         <!-- Auto start -->
         <receiver
             android:name=".receivers.AutoStartReceiver"
--- a/app/src/main/assets/OpenAPSSMB/determine-basal.js
+++ b/app/src/main/assets/OpenAPSSMB/determine-basal.js
@@ -862,8 +862,10 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     console.error("naive_eventualBG:",naive_eventualBG,"bgUndershoot:",bgUndershoot,"zeroTempDuration:",zeroTempDuration,"zeroTempEffect:",zeroTempEffect,"carbsReq:",carbsReq);
     if ( carbsReq >= profile.carbsReqThreshold && minutesAboveThreshold <= 45 ) {
         rT.carbsReq = carbsReq;
+        rT.carbsReqWithin = minutesAboveThreshold;
         rT.reason += carbsReq + " add'l carbs req w/in " + minutesAboveThreshold + "m; ";
     }
+
     // don't low glucose suspend if IOB is already super negative and BG is rising faster than predicted
     if (bg < threshold && iob_data.iob < -profile.current_basal*20/60 && minDelta > 0 && minDelta > expectedDelta) {
         rT.reason += "IOB "+iob_data.iob+" < " + round(-profile.current_basal*20/60,2);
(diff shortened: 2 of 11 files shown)


MilosKozak:
@AdrianLxM this is not part of https://github.com/MilosKozak/AndroidAPS/pull/1950 (latest master). Do we allow this change.
Basicaly the idea is good.
How is this solved in OAPS?

twain47:
In OpenAPS this information is sent to the user using a pushover notification

twain47:
In OAPS the full text of rT.reason is sent to the user if carbsReq > 0 (which is returned) however that seemed a bit verbose in this case.

AdrianLxM:
A minor change: Once it is carbs*Required* and once it is carbs*Requested*.

Also I'm not sure if we should use the openloop notifications or the standard ones?

MilosKozak:
i'd say separate notification. it's usefull in closed loop as well

twain47:
I would have liked to have made the carbs suggestions appear at the bottom of the overview screen  in the same style as open loop suggestions but I couldn't work out how. Would be very grateful if someone could point me at how that works and so I can make that change.

MilosKozak:
open loop button requires special handling.
We could display it next COB and create notifications in urgent cases

jotomo:
@twain47 OpenAPS has the _carbsReqThreshold_ (see _SMBDefaults_) to avoid triggering this too frequently, which users coming from OpenAPS are probably gonna request next. Furthermore, I'd expected more user requests for more configurability (sound, override silent, snooze ...).  For alerts there probably needs to be something like snoozing once the alarm is confirmed, since (in my observations) _carbsReq_ can stay up until an upward trend is observed, despite carbs being administered. 
The point I'm trying to make: just fetching the data and displaying it is just the start. Ideally and IMHO the whole idea deserves to be fleshed out (and dare I say designed) before being implemented and pushed to users to accommodate real-life use cases including alarm fatigue(!) and an option to temporarily silence the alert (this applies to existing local alerts for connection issues as well).
That being said, hooray for a carbs required alert!

twain47:
@jotomo Adding carbsReqThreshold seems reasonable. We always ran OpenAPS with it set to 1 so got the expected behaviour. I'll try and get that added later.

The snooze, etc. don't seem to be much of an issue for us at the moment (we have been running with something like this patch for about 3 weeks). It hasn't been my experience that it can stay on after an upwards trend - possibly a difference in how AAPS triggers the loop? At the moment the alert is very (overly?) subtle which means most the the other issues don't really cause a problem. The main issue is missing the alert altogether.

However, the main issue, I'm not an experienced android developer. Adding alert options, dismiss, etc. is probably outside my current ability.

Tornado-Tim:
I have been running this, for about a day or so, and havnt found the alarms to be annoying, infact I have been more appreciative of the extra warning.

But yes, the ability to snooze and have a threshold for it to trigger the alert first is a good idea.

wootmasterslick:
Its been a while. What is the current status of this? Love to see it in AAPS.

twain47:
> Its been a while. What is the current status of this? Would love to see it in AAPS.

Yes, you're right. I'll update the PR as per feedback.

I got involved in other things and didn't come back to it.

twain47:
Screenshot (10 Mar 2020 12_21_01)

Added next to COB (as suggested) but space is a bit tight for meaningful text

guydavies:
Hi Brian,

That looks cool.  Why not abbreviate to (+21g in 44min)?  You could even remove the + because it relates to COB and you only show requirements in brackets.

Rgds,

Guy

wootmasterslick:
Abbreviating it might make it harder to understand for less technical users, so
if there is enough space I think you could leave it as is.  Looks very good! 
Is there also a notification triggered if the prediction becomes “critical” ?

twain47:
> Is there also a notification triggered if the prediction becomes “critical” ?

There is a notification and you can specify the threshold (number of carbs) before it is triggered.

twain47:
@AdrianLxM @MilosKozak 
I think this is updated based on current feedback and rebased onto dagger3.

Tornado-Tim:
@twain47 can you change the base to dev? all merging work will now be done on that branch.

PR #1975, opened 2019-08-21
[Testing] Location start foreground

jotomo:


MilosKozak:
i used this approach for DummyService (calling in onStart and onCreate) and it seemed to helped.
Do you know how to identify all places where is it needed?

jotomo:
@MilosKozak From what I've read (the official docs aren't really clear on this unfortunately), the best bet is to have the call in _onCreate_ (not always invoked, but if it is, this is the quickest to make the required call) and _onStartCommand_. Invoking _startForegroundService_ multiple times does no harm. Calling _stopSelf_ before invoking _startForegroundService_ would cause a crash, but that method is only invoked when user requests app shutdown.
I've checked the code and _LocationService_ was the only place not following the above, so I updated it and was hoping this would solve it. But @parapenT1sta reported it still crashes. I've added the bit that logs the exception, let's see if it is still the same exception when he tests again.

MilosKozak:
how did you find out that the issue should be LocationService?

jotomo:
I just looked at invocations of _Service.startForeground_ and then checked the services called by it (starting from this exception https://github.com/MilosKozak/AndroidAPS/pulls#issuecomment-523506141)

parapenT1sta:
Now I am able to run AndroidAPS :) Even with some other errors at the startup that I never had before. I shared them on the Issue created on GitHub.

THANK YOU SO MUCH!

MilosKozak:
last commit seems to make it worse
i created branch locationforeground

jotomo:
@MilosKozak Last commit broke updates to persistent notification as collateral damage. The commit bundled my remaining ideas to fix the crash (I'm suspecting _PersistentNotification.getLastNotification returning a dummy notification to be cause_; _AlarmSoundService_ calls _startForeground_ but if the passed notification is a dummy, it probably doesn't count. I've been doing some refactoring to make sure an ongoing service and notification are available reliably at start (https://github.com/MilosKozak/AndroidAPS/compare/dev...jotomo:ongoing-notification?expand=1), but I'm afraid I'm messing up other scenarios that I can test ...

jotomo:
@MilosKozak ... removing _PersistentNotificationPlugin.getLastNotification_ method and call from _AlarmSoundService_ and invoke _startForegroundService_ like in other places?

PR #1977, opened 2019-08-22
Log an uncaught exception before it crashes the app.

jotomo:
According to https://fabric.io/kits/android/crashlytics/features (Step 3, 2nd paragraph), registering an UncaughtExceptionHandler before Fabric is set up does not interfere with Fabric.

This will help getting access to stracktraces (via the log) which otherwise are only visible via logcat.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/MainApp.java
+++ b/app/src/main/java/info/nightscout/androidaps/MainApp.java
@@ -139,6 +139,8 @@ public class MainApp extends Application {
         sConstraintsChecker = new ConstraintChecker();
         sDatabaseHelper = OpenHelperManager.getHelper(sInstance, DatabaseHelper.class);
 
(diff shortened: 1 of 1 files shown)


MilosKozak:
should not we send it to firebase by calling `FabricPrivacy.logException(...)`?

jotomo:
Fabric already does that, it also registers an uncaught exception handler. However, the issues with starting foreground services didn't create Fabric reports. (When Android asks the JVM to shut down, it disallows creating new threads. But since Fabric creates network traffic it either must have a thread on standby or create a new background thread since network I/O on the main thread is not allowed. So my speculation is that Fabric is unable to report in that scenario due to this).
In short: this handles the case where Fabric doesn't report anything but makes sure something goes into the AAPS' logs, rather than just Android's system log.

MilosKozak:
great!

MilosKozak:
let's try it

PR #1982, opened 2019-08-24
LGS watchdog

MilosKozak:
The idea is to force at least some basal if APS is requesting zerotemp for a too long time to prevent ketoacidosis

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/MainApp.java
+++ b/app/src/main/java/info/nightscout/androidaps/MainApp.java
@@ -41,6 +41,7 @@ import info.nightscout.androidaps.plugins.aps.openAPSMA.OpenAPSMAPlugin;
 import info.nightscout.androidaps.plugins.aps.openAPSSMB.OpenAPSSMBPlugin;
 import info.nightscout.androidaps.plugins.configBuilder.ConfigBuilderPlugin;
 import info.nightscout.androidaps.plugins.constraints.dstHelper.DstHelperPlugin;
+import info.nightscout.androidaps.plugins.constraints.lgsWatchdog.LgsWatchdogPlugin;
 import info.nightscout.androidaps.plugins.constraints.objectives.ObjectivesPlugin;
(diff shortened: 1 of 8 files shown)


scottleibrand:
I have two substantial concerns with this. 

First, without an IOB check, this would limit AAPS’ ability to zero-temp to mitigate the low BG from the residual insulin activity from a big bolus. For example, if you have a 1:10 CR and do a 6U bolus for 60 carbs, but end up not being able to eat them, IOB and insulin activity will likely still be positive after 2.5h of zero temping, and raising basal from zero to 10% would slow down AAPS’ ability to counteract the big bolus and normalize BG. 

Secondly, euDKA is only likely after 12-24h of fasting. The proper reaction to a too-long period of low BG despite <10% insulin delivery is to alert the user that they need to eat something. We should present an alarm to the user within AAPS to that effect, preferably before we force resumption of insulin delivery (in spite of low BG).

swissalpine:
I like this idea .,. and I have a selfmade addition to determinebasalresultsmb.java (43ff.) which works really good for me. Perhaps something to think about: I check if one hour br is missing and give if it is the case 20% tbr.
```
TreatmentsPlugin.getPlugin().updateTotalIOBTreatments();
                TreatmentsPlugin.getPlugin().updateTotalIOBTempBasals();
                final IobTotal bolusIob = TreatmentsPlugin.getPlugin().getLastCalculationTreatments();
                final IobTotal basalIob = TreatmentsPlugin.getPlugin().getLastCalculationTempBasals();
                // Anpassung Basalrate
                double baseBasalRate = info.nightscout.androidaps.plugins.configBuilder.ConfigBuilderPlugin.getPlugin().getActivePump().getBaseBasalRate();
                if ((bolusIob.iob + basalIob.basaliob) < (0 - baseBasalRate)) {
                    double cutoff = baseBasalRate * 0.2;
                    if (rate < cutoff) rate = cutoff;
                }
```

PR #2027, opened 2019-09-05
Add extended statuslight to overview fragment

triplem:
extend the already existing statuslight using concrete values, so that an additional look on the CP or the Pump page is not necessary.


MilosKozak:
do you have screenshot?

triplem:
Right now the colouring is not working correctly.

Screenshot

The old statusline can still be used, it is a separate switch button. Unfortunately right now, you even can select both (the old as well as the new statusline) can be selected, but then only the old one is shown.

triplem:
See Preferences Screenshot. This shows, that there are now two switch buttons for selecting the wanted statusline. Like stated, it is possible to select both, but only the first one is shown on the overview fragment. 

Screenshots Prefs

PR #2092, opened 2019-10-04
TriggerPumpDisconnected: Initial work

PoweRGbg:
copy&paste from TriggerBolusAgo


MilosKozak:
maybe PumpLastConnection would be better name for this

PR #2094, opened 2019-10-05
Enable SMB for High BGs

stevebell117:
So one thing I've wanted out of *APS for awhile is the ability to turn on SMBs only when a high BG is detected. This PR is aimed at accomplishing that, as well as a refactor to how the SMB logic is structured because it was a pain trying to work in the existing structures. The re-design should not have changed the existing behavior.

I have done some personal testing, and it works when I have a BG greater than the Top of the range I have set via NSClient. I have not yet tested Simple profiles or the other, but I wanted to get some initial eyes on this for concerns/changes before I go too deep into testing.

Let me know if you need anymore info!

### DIFF
--- a/app/src/main/assets/OpenAPSSMB/determine-basal.js
+++ b/app/src/main/assets/OpenAPSSMB/determine-basal.js
@@ -58,7 +58,7 @@ var enable_smb = function enable_smb(
 ) {
     // disable SMB when a high temptarget is set
     if (! microBolusAllowed) {
-        console.error("SMB disabled (!microBolusAllowed)")
+        console.error("SMB disabled (!microBolusAllowed)");
         return false;
     } else if (! profile.allowSMB_with_high_temptarget && profile.temptargetSet && target_bg > 100) {
         console.error("SMB disabled due to high temptarget of",target_bg);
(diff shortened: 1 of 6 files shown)


app/src/main/assets/OpenAPSSMB/determine-basal.js
```diff
@@ -48,6 +48,83 @@ function convert_bg(value, profile)
...
+            console.error("Warning: SMB enabled within 6h of using Bolus Wizard: be sure to easy bolus 30s before using Bolus Wizard");
+        } else {
+            console.error("SMB enabled for temptarget of",convert_bg(target_bg, profile));
+        }
+        return true;
+    } 
+    
+    // enable SMB if high bg is found
+    if (profile.enableSMB_high_bg === true && high_bg !== null && bg >= high_bg) {
+        console.error("Checking BG to see if High for SMB enablement.");
```
stevebell117: This log doesn't make sense anymore, will need to be removed.

MilosKozak:
sorry. we don't modify determinebasal at all. it would lead to huge problems when updating to new versions. if you want to do changes, create PR to oref0 repo. Then we'll use it on next determinebasal update

PR #2106, opened 2019-10-09
Save Dexcom Sensor Type

TebbeUbben:


MilosKozak:
is OK with this? https://github.com/MilosKozak/AndroidAPS/blob/dev/app/src/main/java/info/nightscout/androidaps/plugins/source/SourceNSClientPlugin.java#L99
@TebbeUbben

PR #2130, opened 2019-10-22
Work implementation of Adroid.BackupAgent

PoweRGbg:
Not completely tested

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -31,6 +31,7 @@
 
     <application
         android:name=".MainApp"
+        android:backupAgent="info.nightscout.androidaps.utils.aBackupAgent"
         android:allowBackup="true"
         android:icon="${appIcon}"
         android:label="@string/app_name"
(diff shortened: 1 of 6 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -31,13 +31,18 @@
...
         android:icon="${appIcon}"
         android:label="@string/app_name"
         android:roundIcon="${appIconRound}"
         android:supportsRtl="true"
         android:theme="@style/AppTheme.Launcher"
         android:fullBackupContent="true">
+
+        <meta-data android:name="com.google.android.backup.api_key"
+            android:value="AEdPqrEAAAAI3JiApyMrbP2QFzZ2fYfCPsgjkRp53Dm2S1-zPQ" />
+
```
MilosKozak: where is the key from?

PoweRGbg: https://developer.android.com/google/backup/signup.html

PoweRGbg:
Unefficient !

Issue #1304, opened 2018-08-05
[Improvement] Silent Alarms


chriske64:
I'm using the RC6.1 with a combo pump. Sometimes I get alarms (eg. Pump unreachable, or pump history changed) during meetings. It is very unpleasant because it attracts all the attention and I always need to explain myself.
To overcome this problem I have put my phone into 'total silence' mode which supresses any sounds. But then I forgot to switch it off again and missed my waking up alarm. 

I would like to have a feature to completely turn off all sounds made by the app. Needless to say I think it should be not a default setting but I would really appreciate such a setting.

Thanks!

triplem:
Hm, you state that you forget to turn on the alarms again. This happens to me as well, and I really have it especially for AAPS. Furthermore some alarms, at least for me, are important, even while sitting in a meeting with eg a customer (urgent low). I guess we should allow to customize urgent alarms and "normal" alarms differently. What do you think?

chriske64:
Personally, I have muted the low/high alarms anyway. I can feel it etc.
I think you raise a good point but for me the most important part is that I can have a setting where the app does not make any sounds.

andyrozman:
Yes silent mode is something that I would really need... Sometimes you just don't want to hear the alarms (important meetings, etc). 

Solution would actually be pretty simple... If user clicks on "Silent mode" all current alarm setting would be saved and all alarms turned off, and when user would select "Normal mode" again, all previous alarm settings would be restored.

jroth-haj:
So far, as I know, this alarms are played as media. I also use dnd mode. The alarms are still coming up.
Because of that I have the media volume at 0.
When I hear music, I must raise up this.
When I forgot to raise down it after listening to music, I get loud alarms in urgent meetings. :-(

Why the app can't use system alarms?
The alarms also should be classified.

lee-b:
I'm willing to (going to, if only for personal use) implement this "silent mode" (or something equivalent).  For one thing, the alarms were so distracting on a motorbike lately that it was dangerous.  I also have the work problem, like others mentioned above.

I'm new to android development, java development, and this codebase, but not dev more generally.  Any tips on where to start this?

I'm thinking it would be nice to implement this through the same notificationbuilder that other parts of the codebase (seem to) use for alerts, which support vibration instead of just sounds.  Don't really get why these particular alerts just play sounds directly.

However, as a start, I might just add a "silent mode" config option that overrides these particular alerts.

lee-b:
[...]

PR #2133, opened 2019-10-23
Setting to play alerts as notification / 100% music stream

anbraten:
Having error sounds played as media was quite unhandy (esp when you are in a call). I changed the audio stream to notification, so you can set you phone in DND mode and be able to leave your media volume untouched. A media stream should IMHO only be used for foreground activities / actions (playing music / watching a movie).

#1304


MilosKozak:
@AdrianLxM does this have any side effects?

AdrianLxM:
Hi @anbraten,

thanks for your PR! It indeed seems like the proper way to not use the `AudioManager. STREAM_MUSIC` volume.

It would affect the behaviour that people are used to with alarms (e.g. having AAPS alarms at night while still being on silent for phone calls).

The possibility to have this for one app (or just one notification type of an app) exists on newer phones - but deeply hidden in notification setting sub-menus and only accessible if you click on items that don't even look actionable.

We should therefore have this feature go through proper beta/RC stages. Sorry that we will only merge it after the upcoming release.

We should then have a simple "override silentmode" option for the user.

Did you test the behaviour of `AudioManager.STREAM_ALARM` volume?
If that would also meet your needs - I prefer going with default Android configurability.

anbraten:
Hey @AdrianLxM,

> It would affect the behaviour that people are used to with alarms (e.g. having AAPS alarms at night while still being on silent for phone calls).

It seems there are two cases:

- AAPS should not disturb (eg. meeting, driving a car, theater, ...)
- AAPS should even disturb when phone normally in silent mode (eg. sleeping)

> The possibility to have this for one app (or just one notification type of an app) exists on newer phones - but deeply hidden in notification setting sub-menus and only accessible if you click on items that don't even look actionable.
> We should then have a simple "override silentmode" option for the user.

I agree with you, AAPS should not depend on some special functions which currently aren't implemented by most manufactures at all.

> Did you test the behaviour of AudioManager.STREAM_ALARM volume?
> If that would also meet your needs - I prefer going with default Android configurability.

I tried `STREAM_ALARM` as well, but the problems is still that I am scared about unexpected notifications in some situations. ;) My clock alarms are only set on special times at which I am normally not working.

Right now I would suggest  we use `STREAM_NOTIFICATIONS` as suggest in my PR as default and add an option "override silent mode" which will set it to `STREAM_MUSIC` with full volume.

In the long term I will maybe create a PR of a notification rework based on the `persistentNotification`, so you could end an alarm / mute it like you can do in xdrip+.

anbraten:
I added an option to turn off the `Play alerts with 100% volume as music stream` setting. It is normally enabled to keep the known behavior.

MilosKozak:
@AdrianLxM I'd prefen not adding more options. What is the common way to do it ? How is it done in Dexcom app?

wootmasterslick:
You can have a look in the old sightremote app from TebbeUbben. I believe pump errors were handled quite well there.

EknalpTiel:
So you find a solution to geht this in Version 2.6?

PR #2155, opened 2019-10-29
Rewrite PointsWithLabelGraphSeries

AlmazKo:
Here are some improvements in PointsWithLabelGraphSeries class without any changes in rendered result.

I added internal Metrics class and changed slightly the class structure.

PS: And I would be happy to continue improving the graph


app/src/main/java/info/nightscout/androidaps/plugins/general/overview/graphExtensions/PointsWithLabelGraphSeries.kt
```diff
@@ -0,0 +1,419 @@
...
+
+    companion object {
+        /**
+         * helper to render triangle
+         *
+         * @param point array with 3 coordinates
+         * @param canvas canvas to draw on
+         * @param paint paint object
+         */
+        private fun drawArrows(point: Array<Point>, canvas: Canvas, paint: Paint) {
```
MilosKozak: is it necessary to use companion object here?

PR #2172, opened 2019-11-01
Add possibility to show Battery info in days instead of percent

gregorybel:
Battery info for Combo in Percent is not so helpful.
Better is to see the numbers of battery days, So I added an option to enable this.

Before:

After:

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/StatuslightHandler.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/StatuslightHandler.java
@@ -12,6 +12,7 @@ import info.nightscout.androidaps.interfaces.PumpInterface;
 import info.nightscout.androidaps.plugins.configBuilder.ConfigBuilderPlugin;
 import info.nightscout.androidaps.plugins.general.careportal.CareportalFragment;
 import info.nightscout.androidaps.plugins.general.nsclient.data.NSSettingsStatus;
+import info.nightscout.androidaps.plugins.pump.common.defs.PumpType;
 import info.nightscout.androidaps.utils.DecimalFormatter;
 import info.nightscout.androidaps.utils.SP;
 import info.nightscout.androidaps.utils.SetWarnColor;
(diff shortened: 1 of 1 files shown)


MilosKozak:
I'm OK with the idea but i don't want add more options. Do it based on pump plugin used

gregorybel:
ok, I can do that but what do you suggest if some users anyway prefer seeing battery percentage? How to handle that?

jotomo:
> ok, I can do that but what do you suggest if some users anyway prefer seeing battery percentage? How to handle that?

Don't handle it at all. Battery percentage is synthetic for the Combo, it serves little value beyond informing about a low or empty battery and that information is already conveyed via the colour.
Though the Combo plugin should return a value <= 22 for low battery, rather than the 25 it's currently returning to work with the default warning level threshold of 22:

https://github.com/MilosKozak/AndroidAPS/blob/0bec3cb224815ae53d0974c2e67be5577984a0b1/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/StatuslightHandler.java#L39

https://github.com/MilosKozak/AndroidAPS/blob/6044335b1d754208c24b1a8be3d88c303d8e2bcc/app/src/main/java/info/nightscout/androidaps/plugins/pump/combo/ComboPlugin.java#L429

gregorybel:
@MilosKozak @jotomo done, is it ok like this?

MilosKozak:
👍  
@jotomo ?

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/StatuslightHandler.java
```diff
@@ -34,10 +35,14 @@ void statuslight(TextView cageView, TextView iageView, TextView reservoirView,
...
-        applyStatuslightLevel(R.string.key_statuslights_bat_critical, 5.0,
-                R.string.key_statuslights_bat_warning, 22.0,
-                batteryView, "BAT", batteryLevel);
+        if (pump.model() != PumpType.AccuChekCombo) {
+            double batteryLevel = pump.isInitialized() ? pump.getBatteryLevel() : -1;
+            applyStatuslightLevel(R.string.key_statuslights_bat_critical, 5.0,
+                    R.string.key_statuslights_bat_warning, 22.0,
+                    batteryView, "BAT", batteryLevel);
+        } else {
+            applyStatuslight("page", CareportalEvent.PUMPBATTERYCHANGE, batteryView, "BAT", 504, 240);
```
jotomo: "page"?

gregorybel:
push wrong button, yes good catch, will change this!

gregorybel:
@MilosKozak can we merge or anything else?

PR #2178, opened 2019-11-03
Countdown time for temporary PS in overview

2flea:
I added a countdown time for temporary PS in overview analog to countdown time for TT. Especially for PS initiated by Automation plugin this is very helpful as you can see for how long the PS is scheduled.

Pocophone F1:
image
Nexus 4:
image
Atom:
image

I did a modification in OverviewFragment to check for PS with duration when retrieving profile name (needed to import PS class).
Please let me know if you'd rather have this encapsulated in a ProfileFunction to retrieve proper profile name?!

This is my 1st PR, please tell me if something is not what you expect, so I can learn :-)


MilosKozak:
Instead of computing it in fragment add function 
`ProfileFunctions.getInstance().getProfileNameWithDuration()`

and set it here

https://github.com/MilosKozak/AndroidAPS/pull/2178/files#diff-7b8b3a371d7af16bc8a131e8c7d87c1dR1228

2flea:
@MilosKozak I pushed it to my fork, I will need to test it first. Did not know that it automatically will pop up here in the PR. 
I will let you know as soon as I finished my tests. 
Andreas

2flea:
Did some tests at home with my spare phone. Looking good.

PR #2198, opened 2019-11-12
Log unhandled exceptions again

jotomo:


MilosKozak:
shouldn't we use log.error in this case?

jotomo:
> shouldn't we use log.error in this case?

Indeed. PR updated.

PR #2199, opened 2019-11-12
WIP: Automation location left trigger

anbraten:
It would be nice to trigger on location left in automation as well as on location entered.

This way you could say: After leaving work, I want a special TT while cycling home and at home it should be stopped again.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/general/automation/triggers/TriggerLocation.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/general/automation/triggers/TriggerLocation.java
@@ -13,12 +13,16 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.text.DecimalFormat;
+import java.util.ArrayList;
+import java.util.Arrays;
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/general/automation/triggers/TriggerLocation.java
```diff
@@ -106,6 +121,7 @@ Trigger fromJSON(String data) {
             longitude.setValue(JsonHelper.safeGetDouble(d, "longitude"));
             distance.setValue(JsonHelper.safeGetDouble(d, "distance"));
             name.setValue(JsonHelper.safeGetString(d, "name"));
+            mode.setValue(JsonHelper.safeGetString(d, "mode"));
```
MilosKozak: as thi is upgrading current class you must provide default value here

anbraten: added location entered as default value

MilosKozak:
I think it will not work when you switch language. You have to store to JSON constants defined in code

PR #2206, opened 2019-11-17
Correct typo in TT start string (DE)

2flea:
Tempörär --> Temporär


MilosKozak:
this has to be changed through Crowdin

2flea:
Milos, I checked crowdin but there was no such string. Can you explain what is required to add it?

Issue #2215, opened 2019-11-22
Profils witch with deactivated loop


DanielaBose:
In the Karlsruher group is a member who has just noticed that the Automation Plugin executes a profile switch even though the loop is deactivated.
AAPS 2.5.1 Master

MilosKozak:
resolved in dev

PR #2216, opened 2019-11-22
Deactivate Automation if Loop is disabled, not only if it is suspended

2flea:
Currently Automation is only deactivated if Loop is suspended for a temporary time (1h, 2h, 3h). 
Check also if Loop Plugin is disabled. (Loop deactivated in Overview)

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/general/automation/AutomationPlugin.kt
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/general/automation/AutomationPlugin.kt
@@ -161,7 +161,7 @@ object AutomationPlugin : PluginBase(PluginDescription()
     private fun processActions() {
         if (!isEnabled(PluginType.GENERAL))
             return
-        if (LoopPlugin.getPlugin().isSuspended) {
+        if (LoopPlugin.getPlugin().isSuspended || !LoopPlugin.getPlugin().isEnabled(PluginType.LOOP)) {
             if (L.isEnabled(L.AUTOMATION))
                 log.debug("Loop deactivated")
             return


2flea:
Fix for issue #2215

AdrianLxM:
If you see the Facebook thread in Looped-DE:

People have rules like "Do a profile switch to 130% if my BG is above 130mg/do" that only make sense if the loop is deactivated.
So people seem to use Automation when the loop is deactivated - or instead of the loop algorithm.

We might want to activate/dractivate it fully independently (except for disconnected pumps)?

AdrianLxM:
I agree that the current UX is not optimal.
Another idea would be a pre-condition "loop active" that the user can tick off?
Or just go with Andreas' solution and just use it together with activated loop?
@MilosKozak, what's your opinion?

2flea:
@AdrianLxM currently it is incomplete as it only deactivates Automation for suspended loop but ignores deactivated loop.
If you ask me I do not want to have any automatic behaviour if I switch off (or suspend) loop - I can do PS manually if I want to.

DanielaBose:
I think it's a new and very good feature, but users are definitely assuming that if the loop is disabled, it will not intervene in any way. If it was planned so that the automation can make a profile change also with deactivated loop, then this should, as suggested by Andreas fast be deactivatable.

MasterPlexus:
> I agree that the current UX is not optimal.
> Another idea would be a pre-condition "loop active" that the user can tick off?
> Or just go with Andreas' solution and just use it together with activated loop?
> @MilosKozak, what's your opinion?

I would say, to include the loop status as a must precondition will make the automatic rules very much more clear also from the understanding when they will be set.

MilosKozak:
OK, let's create a trigger loop is enabled and add it as a precondition.
Will you do that?

2flea:
Sorry Milos, I am not sure how to do that.

MilosKozak:
i'll merge for now but we'should change it in the future

PR #2225, opened 2019-11-27
Localprofile improvement, global units

MilosKozak:


app/src/main/java/info/nightscout/androidaps/activities/PreferencesActivity.java
```diff
@@ -96,6 +104,20 @@ private static void updatePrefSummary(Preference pref) {
                     plugin.updatePreferenceSummary(pref);
                 }
             }
+            // convert preferences values to current units
```
jotomo: Feeling the need to write a comment like this (explaining a block) can be a good opportunity to check if code should be in a separate method. Which I'd argue for in this case, since it doesn't just update code for display, but manipulates preference values.
Having it as a separate method like `adjustUnitDependentPrefs` also makes it harder to forget about extending that code when a new pref with a unit is added.
(Originally I was gonna argue to run this code only when the user changes the unit, so it's clearer when the conversion happens, but the 32-separator observation (making the conversion conditional and idempotent) is a clever thought and invoking the code here covers corner cases like a user entering mmol values and only after doing so setting the unit pref correctly to mmol)

MilosKozak: thanks, i'll create new method
I was thinking about specifying affected keys in plugins but finaly affected keys are somehow general and i didn't find better place for them

MilosKozak: @jotomo

PR #2228, opened 2019-11-29
Update Oref to 0.7.0

Tornado-Tim:
This PR bring oref up to date with oref 0.7.0.

Oref 0.7.0 now has the ability to loop from Raw Data, this is not enabled in this PR as only 1 CGM source can provide this. I have added noise/raw data to be pulled in later when we wish to add this ability in later.

This PR enables several previously non-editable preferences:
resistance_lowers_target and sensitivity_raises_target. With the former set default to off and later to on.
By default on upgrade these options will not alter any loop settings, as these settings were already set by default, this just simply enables the presences to be turned off or on.

This PR brings in a whole heap of upgrades to SMB:

UAMSMBBasalMinutes, When UAM is detected the Basal minutes allocated for each SMB can be shifted upwards/downwards from the usual SMBBasal minute. So this allows the use of giving a more hefty bolus to deal with food, and then revert to a more conservative amount when no UAM is detected.

A minimum bolus can now be set for SMB's, depending on the smallest bolus amount your pump can support. For example on a Dana RS the SMB Bolus amount could go down to 0.05 units- thats a very tiny SMB :)

SMB Time - You can now set how often SMB's should be given (Down to 1 min). In theory anything less than 5 min will not be enacted in a automated fashion as CGM values only come in on the 5 minutes, but this would allow SMB's to be eventually dosed more frequently when CGM updates become more frequent.

Carb Suggestion Threshold: This allows oref to restrict when carbs required messages are output from the algorithm. This will be more effective when PR: #1974 is merged in.

I have been running this myself for sometime and works well, however I would appreciate more people testing this.


app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/DetermineBasalAdapterSMBJS.java
```diff
@@ -220,7 +228,8 @@ public void setData(Profile profile,
     ) throws JSONException {
 
         String units = profile.getUnits();
-
+        Double bolusincrement = SP.getDouble("key_bolus_increment", SMBDefaults.bolus_increment);
```
MilosKozak: `Voting for simplicity and using only `pumpbolusstep`

swissalpine:
Thank you for the update.
I've tested it for more than two days now and am convinced: After adjusting the UAMSMBBasalMinutes (120 min) and resetting the SMBBasalMinutes (from 90 to 30 min), SMB works perfectly for me for the first time (at least for the past two days). 
I would be happy about an official update to oref0.7!

app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/DetermineBasalAdapterSMBJS.java
```diff
@@ -236,15 +245,13 @@ public void setData(Profile profile,
         mProfile.put("max_daily_safety_multiplier", SP.getInt(R.string.key_openapsama_max_daily_safety_multiplier, 3));
         mProfile.put("current_basal_safety_multiplier", SP.getDouble(R.string.key_openapsama_current_basal_safety_multiplier, 4d));
 
-        // TODO AS-FIX
-        // mProfile.put("high_temptarget_raises_sensitivity", SP.getBoolean(R.string.key_high_temptarget_raises_sensitivity, SMBDefaults.high_temptarget_raises_sensitivity));
-        mProfile.put("high_temptarget_raises_sensitivity", false);
-        //mProfile.put("low_temptarget_lowers_sensitivity", SP.getBoolean(R.string.key_low_temptarget_lowers_sensitivity, SMBDefaults.low_temptarget_lowers_sensitivity));
-        mProfile.put("low_temptarget_lowers_sensitivity", false);
```
jotomo: Those options are disabled due to an issue where AS produced some extreme values. I don't think that issues has been fixed so far.

app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/DetermineBasalAdapterSMBJS.java
```diff
@@ -326,6 +341,8 @@ public void setData(Profile profile,
         mMicrobolusAllowed = microBolusAllowed;
         mSMBAlwaysAllowed = advancedFiltering;
 
+        mCurrentTime = System.currentTimeMillis();
```
jotomo: Also use _now_ here, like above?

app/src/main/java/info/nightscout/androidaps/plugins/iob/iobCobCalculator/GlucoseStatus.java
```diff
@@ -40,6 +42,7 @@ public GlucoseStatus() {
 
     public GlucoseStatus round() {
         this.glucose = Round.roundTo(this.glucose, 0.1);
+        this.noise = Round.roundTo(this.noise, 0.1);
```
jotomo: The noise calculation coming in oref0 0.7.1 needs two decimal places.

Tornado-Tim:
Hi All I have fixed the min SMB amount to the min supported by the pump instead of it being configurable for simplicity.
Fixed issues mentioned in review.

stevebell117:
I have been running this branch for 4-6 days, and have not run into any noticeable issues.

Mackwe:
I've been running this since mid December at least. Fantastic results. No issues at all.

rhys-jones:
Working well for me for a couple of weeks, no issues.

stevebell117:
So this may not be an issue with this branch, but AAPS dev as a whole, but I've noticed that comms to my pump break if there is a consistent Delta of +10 with no user inputted insulin. If a COB is entered, but no insulin, this also still breaks it and both Temp Basals and SMBs are not given.

I will provide the logs once I get home and parse them, but the algorithm is submitting faulty data (a bunch of 00 bytes) and the pump is NOT ACKing the requests. The only way to resume comms and enable loops it to push a manual bolus through AAPS.

Once again, will update this when I get a chance to parse the logs, but wanted to get it on your radar @Tornado-Tim

stevebell117:
[broken basal_smb logs.zip](https://github.com/MilosKozak/AndroidAPS/files/4115309/broken.basal_smb.logs.zip)
Here are the logs for the aforementioned issue I mentioned. Please let me know if you need anymore information.

Tornado-Tim:
> So this may not be an issue with this branch, but AAPS dev as a whole, but I've noticed that comms to my Medtronic 523 pump break if there is a consistent Delta of +10 with no user inputted insulin. If a COB is entered, but no insulin, this also still breaks it and both Temp Basals and SMBs are not given.
> 
> I will provide the logs once I get home and parse them, but the algorithm is submitting faulty data (a bunch of 00 bytes) and the pump is NOT ACKing the requests. The only way to resume comms and enable loops it to push a manual bolus through AAPS.
> 
> Once again, will update this when I get a chance to parse the logs, but wanted to get it on your radar @Tornado-Tim

I am not entirely sure this is an issue caused by the PR itself or the upstream code based on the dev branch.

The PR is a bit behind dev now, so I will update the PR again with all the latest commits to dev.

stevebell117:
I am using Dev as of a few days ago with this change merged into it.

stevebell117:
I think I figured out what causes the problem. If you have Max Basal per hour set in AAPS higher than what's allowable by the pump, the pump rejects the command and then the SMB that follows it is ignored.

The behavior happened again tonight, and I noticed it was trying to put a basal .05u higher than my pump allowed which explains the NAK.

Tornado-Tim:
> I think I figured out what causes the problem. If you have Max Basal per hour set in AAPS higher than what's allowable by the pump, the pump rejects the command and then the SMB that follows it is ignored.
> 
> The behavior happened again tonight, and I noticed it was trying to put a basal .05u higher than my pump allowed which explains the NAK.

Hi Steve,

Can you make an issue on github please. The issue doesn't pertain to this fork solely.

alternateal:
I have been running this merged into 2.6 for the last 3 days, and for me it has been working really well. I am using pretty conservative settings 45 mins basal for SMB and 30 mins for UAM. I think it has coped much better than the existing version. I will continue to run with this but would be more than happy if merged into the main branch.

app/src/main/res/values/strings.xml
```diff
@@ -858,8 +858,15 @@
     <string name="bgsource_upload">BG upload settings</string>
     <string name="wear_detailed_delta_title">Show detailed delta</string>
     <string name="wear_detailed_delta_summary">Show delta with one more decimal place</string>
+    <string name="key_smbinterval" translatable="false">key_smbinterval</string>
```
MilosKozak, line 861: should be
`<string name="key_smbinterval" translatable="false">smbinterval</string>`

Tornado-Tim: fixed in latest commit

Issue #2188, opened 2019-11-07
Incorrect basal rate on App compared to pump. Intermittent fault.


DrVoles:
On 28th October 2019 at 16.12 GMT.
AndroidAPS 2.5
Accu chek insight pump.
Dexcom patched app 2.4 region 1 mmol/l
Pixel 3XL Android 10.

Was noted that glucose was dropping. APS advised 0% basal rate, which was shown on the pump as having been enacted - but on APSaon screen - was not showing correct basal. Could this be leading to inaccurate calculations?
Following discussions on Facebook - was advised to change setting: Always use basal absolute values, from on to off.
This appears to have rectified the issue - but means information from autotune is not valid.
Please see attached photos and log.
![IMG_20191028_161210](https://user-images.githubusercontent.com/42710627/68386894-ff47d600-0154-11ea-9b7c-b0398fdd2748.jpg)
![Screenshot_20191028-162211](https://user-images.githubusercontent.com/42710627/68387026-403fea80-0155-11ea-9dc5-6256546ebaae.png)
[AndroidAPS.log](https://github.com/MilosKozak/AndroidAPS/files/3819438/AndroidAPS.log)

Many thanks for your continued hard work!! X
Mark

MilosKozak:
disable uploading as absolute value in NSClient

DrVoles:
As I've said - I've already done this - but this means information from Autotune is not valid. Thus - the issue is not fixed..

Issue #1765, opened 2019-04-25
Difference between TBR on NS graph and treatment history


Paulusma:
I'm using Insight pump and AAPS 2.1, using extended TBR's

Pump history is as follows:

![image](https://user-images.githubusercontent.com/2260419/56732454-b05cc500-675d-11e9-942f-4c217ca1ed15.png)

This is rendered as follows. The suggested return to basal was enacted on the pump and shows up in the treatment history as well. For some reason NS does not look at duration but at start of the next temp (eg at 20:33:29)?

![image](https://user-images.githubusercontent.com/2260419/56732871-db93e400-675e-11e9-93a8-1c3768c5338e.png)

Paulusma:
BTW: not related to insight security issue, i did not do a DB reset or install a new AAPS version

Paulusma:
Yesterday i upgraded to AAPS 2.3 but the problem persists, see below:
![image](https://user-images.githubusercontent.com/2260419/56848730-a4533d80-68ec-11e9-92ed-cb197a28c37f.png)
![Screenshot_20190427-123835](https://user-images.githubusercontent.com/2260419/56848753-cb117400-68ec-11e9-8b4f-069f63addff6.png)
![Screenshot_20190427-123857](https://user-images.githubusercontent.com/2260419/56848754-ccdb3780-68ec-11e9-889e-2bef2c7ba346.png)
![image](https://user-images.githubusercontent.com/2260419/56848763-ef6d5080-68ec-11e9-8855-b7eb7b013dff.png)

MilosKozak:
you have to resolve first so many TBRs starting at the same time

PR #2249, opened 2019-12-07
ensure end of tbr doesn't override next tbr, closes #1963

### LINKED ISSUE #1963: [Insight][2.4-Alpha3] Loop sometimes sets temporary basal and then immediately cancels it
[AndroidAPS_LOG_1565873945457.log.zip](https://github.com/MilosKozak/AndroidAPS/files/3506406/AndroidAPS_LOG_1565873945457.log.zip)

DigitalDan1:
I can see that @TebbeUbben has made changes (in another branch) to the way that end of TBRs are processed from Insight pumps, but I wanted this change in my dev branch to work around the bug for now.  I thought it may be useful for others that want the temporary fix too.

closes #1963

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
+++ b/app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
@@ -898,7 +898,7 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
                 PreparedQuery<TemporaryBasal> preparedQuery2 = queryBuilder2.prepare();
                 List<TemporaryBasal> trList2 = getDaoTemporaryBasal().query(preparedQuery2);
 
-                if (trList2.size() > 0) {
+                if (trList2.size() > 0 && trList2.get(0).pumpId == 0) { // don't update existing record if it has a pumpId
                     old = trList2.get(0);
 
                     old.copyFromPump(tempBasal);
@@ -1001,7 +1001,8 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
         try {
             List<TemporaryBasal> tempbasals;
             QueryBuilder<TemporaryBasal, Long> queryBuilder = getDaoTemporaryBasal().queryBuilder();
-            queryBuilder.orderBy("date", ascending);
+            // order by date and duration to ensure end of previous tbr doesn't override next tbr
+            queryBuilder.orderBy("date", ascending).orderBy("durationInMinutes", ascending);
             Where where = queryBuilder.where();
             where.ge("date", mills);
             PreparedQuery<TemporaryBasal> preparedQuery = queryBuilder.prepare();
@@ -1017,7 +1018,8 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
         try {
             List<TemporaryBasal> tempbasals;
             QueryBuilder<TemporaryBasal, Long> queryBuilder = getDaoTemporaryBasal().queryBuilder();
-            queryBuilder.orderBy("date", ascending);
+            // order by date and duration to ensure end of previous tbr doesn't override next tbr
+            queryBuilder.orderBy("date", ascending).orderBy("durationInMinutes", ascending);
             Where where = queryBuilder.where();
             where.between("date", from, to);
             PreparedQuery<TemporaryBasal> preparedQuery = queryBuilder.prepare();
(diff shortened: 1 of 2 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1190,7 +1190,8 @@ private void processHistoryEvents(String serial, List<HistoryEvent> historyEvent
             temporaryBasal.isAbsolute = false;
             temporaryBasals.add(temporaryBasal);
         }
-        Collections.sort(temporaryBasals, (o1, o2) -> (int) (o1.date - o2.date));
+        // order by date and duration to ensure end of previous tbr doesn't override next tbr
+        Collections.sort(temporaryBasals, (o1, o2) -> (int) (o1.date != o2.date ? (o1.date - o2.date) : (o1.durationInMinutes - o2.durationInMinutes)));
```
Suchiman, line 1194: Was `o1.date == o2.date` ever true in this branch? The updates happened because they were equal after rounding up to seconds but this date has millisecond precision and due to https://github.com/MilosKozak/AndroidAPS/pull/2529/files#diff-1bf616b07f370cbb8d226bc1db91c34bR1556 those milliseconds were random.

app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
```diff
@@ -898,7 +898,7 @@ public boolean createOrUpdate(TemporaryBasal tempBasal) {
                 PreparedQuery<TemporaryBasal> preparedQuery2 = queryBuilder2.prepare();
                 List<TemporaryBasal> trList2 = getDaoTemporaryBasal().query(preparedQuery2);
 
-                if (trList2.size() > 0) {
+                if (trList2.size() > 0 && trList2.get(0).pumpId == 0) { // don't update existing record if it has a pumpId
```
Suchiman, line 901: I don't know what the implications of this change are for pumps other than the insight but i would add `|| trList2.get(0).pumpId == temporaryBasal.pumpId` in case we see the same pump event again, in order to not duplicate it in the database.

Suchiman:
Does the data in nightscout look correct with this change? The changes to the sort in LocalInsightPlugin do not look effective to me so it would still potentially upload both records in the wrong order to nightscout. Also i can't judge the implications on this change for bidirectional nightscout sync.

DigitalDan1:
> Does the data in nightscout look correct with this change? The changes to the sort in LocalInsightPlugin do not look effective to me so it would still potentially upload both records in the wrong order to nightscout. Also i can't judge the implications on this change for bidirectional nightscout sync or other pumps.

Thank you very much for your comments.

Yes, the data in nightscout looks correct.  The Temp Basal event without a duration doesn't affect the new Temp Basal event.  The events are uploaded to nightscout with second precision (https://github.com/MilosKozak/AndroidAPS/blob/0f6f344f6307cbd5765e01a2355314aa204e6ec2/app/src/main/java/info/nightscout/androidaps/plugins/general/nsclient/NSUpload.java#L63), and the date comparisons in nightscout are exclusive (e.g. https://github.com/nightscout/cgm-remote-monitor/blob/41ac93f7217b1b7023ec6ad6fc35d29dcf2e4f88/lib/data/ddata.js#L120).

app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
DigitalDan1: That scenario is handled here: https://github.com/MilosKozak/AndroidAPS/blob/0f6f344f6307cbd5765e01a2355314aa204e6ec2/app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java#L869-L880

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
DigitalDan1: It's very unlikely the dates would be the same in this branch.  But, with changes like your fix to remove random milliseconds (I've made the same change) the dates will often be the same, so this ensures the correct order.

DigitalDan1:
Reflecting on this issue some more, the core problem is that the end TBR and start TBR could occur in the same second but this isn't supported by Android APS and its integration with nightscout. 
 There's no explicit model for handling multiple TBRs in the same second.  This can (and does) occur with the Insight pump, which causes the problems we've seen: 
https://github.com/MilosKozak/AndroidAPS/issues/2566, https://github.com/MilosKozak/AndroidAPS/issues/2484, https://github.com/MilosKozak/AndroidAPS/issues/2435, https://github.com/MilosKozak/AndroidAPS/issues/1963, maybe https://github.com/MilosKozak/AndroidAPS/issues/2188 and https://github.com/MilosKozak/AndroidAPS/issues/1765 too.

In lieu of reworking both systems I think a simpler/safer work-around is to ensure end and start TBRs coming from the Insight pump are always at least one second apart by shifting them back/forward by exactly half a second, respectively.

I've updated this PR to use this approach.

PR #2270, opened 2019-12-12
Show profile even if it was switched with percentage

PoweRGbg:
I have a percentage shifted profile and my NSProfile tab shows nothing until I select a profile manually. I added a simple check for "("
and the check to compare with the string before the opening braces


MilosKozak:
what is the issue?

PoweRGbg:
When you switch the profile with percentage or time shift, this action changes the profile name by adding percentage in braces -> "MyProfile" becomes "MyProfile(150%)" which is not recognized as an existing profile and the NSProfile tab is empty.

MilosKozak:
it's not nsprofile anymore 
do i undestandand it well that you have to only select some profile from the list?

PoweRGbg:
Screenshot_1576662936
Screenshot_1576662952
Is the profile  selected and the one shown in NSProfile the same ?

PoweRGbg:
Screenshot_1576663381
Screenshot_1576663389
And look what happens if we don't have a time shift or percentage

PR #2284, opened 2019-12-15
Watch tap action

Philoul:
I started to developp "direct actions from watchfaces" (only AAPSv2 and Steampunk done)...
Actions available in preference menu : Main Menu, Status Menu, Bolus, Wizard, eCarb, TempT, Pump, Loop, CPP, TDD, and None

Zone available : BG, IOB, COB, TIME, DELTA (Chart is "reserved" for scale, and there is a "Background" zone that allways Launch Main Menu)
=> Not all zones are available on all watchfaces (it depends on Chart position and size, and informations of watchface)
=> With that logic, actions is behind one kind of information and can change according to watchface (i.e. Delta can move from Left (Steampunk) to right (Cockpit)...)

we can change the logic with 5 zones customized in preference whatever the watchface : Top, Bottom, Right, Center, Left... With areas slightly adjusted on each watchface according to the size and position of the Chart.


Philoul:
I finally change the logic with 5 zones : Up, Down, Left, Center and Right zone
The chart zone is reserved for time scaling
Then background ("corner" of watchface) allways show main menu.
I comment first coding with BG, IOB, COB... zone (I can do a final commit to remove it)

Philoul:
This PR is also included in "Color improvement" PR : https://github.com/MilosKozak/AndroidAPS/pull/2314

Philoul:
Close this PR (A new one cleaner will be done)

PR #2285, opened 2019-12-16
Local profile ISF precision with decimal place (0.0) and 0.5 steps for mg/dl

2flea:
- typo fix for string "ns_uploadlocalprofile"
- local profile ISF precision with decimal place (0.0) and 0.5 steps for mg/dl

I encountered today that I could not migrate my NS profile to a local profile as for mg/dl the precision was without decimal place. Not sure how many people would have the same problem as I. 

image

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/profile/local/LocalProfileFragment.kt
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/profile/local/LocalProfileFragment.kt
@@ -98,7 +98,7 @@ class LocalProfileFragment : Fragment() {
         TimeListEdit(context, view, R.id.localprofile_ic, "IC", MainApp.gs(R.string.nsprofileview_ic_label), LocalProfilePlugin.currentProfile().ic, null, HardLimits.MINIC, HardLimits.MAXIC, 0.1, DecimalFormat("0.0"), save)
         basalView = TimeListEdit(context, view, R.id.localprofile_basal, "BASAL", MainApp.gs(R.string.nsprofileview_basal_label) + ": " + sumLabel(), LocalProfilePlugin.currentProfile().basal, null, pumpDescription.basalMinimumRate, 10.0, 0.01, DecimalFormat("0.00"), save)
         if (units == Constants.MGDL) {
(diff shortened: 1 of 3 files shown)


MilosKozak:
you cannot migrate means AAP crashed?

2flea:
No AAPS did not crash but I could not enter my ISF values.

2flea:
modified steps back to 1.0 instead of 0.5 but put the possibility of a precision of 0.1

app/src/main/java/info/nightscout/androidaps/plugins/profile/local/LocalProfileFragment.kt
```diff
@@ -98,7 +98,7 @@ class LocalProfileFragment : Fragment() {
         TimeListEdit(context, view, R.id.localprofile_ic, "IC", MainApp.gs(R.string.nsprofileview_ic_label), LocalProfilePlugin.currentProfile().ic, null, HardLimits.MINIC, HardLimits.MAXIC, 0.1, DecimalFormat("0.0"), save)
         basalView = TimeListEdit(context, view, R.id.localprofile_basal, "BASAL", MainApp.gs(R.string.nsprofileview_basal_label) + ": " + sumLabel(), LocalProfilePlugin.currentProfile().basal, null, pumpDescription.basalMinimumRate, 10.0, 0.01, DecimalFormat("0.00"), save)
         if (units == Constants.MGDL) {
-            TimeListEdit(context, view, R.id.localprofile_isf, "ISF", MainApp.gs(R.string.nsprofileview_isf_label), LocalProfilePlugin.currentProfile().isf, null, HardLimits.MINISF, HardLimits.MAXISF, 1.0, DecimalFormat("0"), save)
+            TimeListEdit(context, view, R.id.localprofile_isf, "ISF", MainApp.gs(R.string.nsprofileview_isf_label), LocalProfilePlugin.currentProfile().isf, null, HardLimits.MINISF, HardLimits.MAXISF, 1.0, DecimalFormat("0.0"), save)
```
MilosKozak: this is useless when step is 1.0 or?

2flea: If you use the + and - it will use a step of 1.0 but if you wish to be more precise you can enter manually e.g. 53.7. 
NS allows precision of 0.x, I assumed local would do the same. But never mind if not I will just adapt my ISF to be full numbers.

MilosKozak:
ok

PR #2305, opened 2019-12-23
refactor overview with new header area and better formated user value…

osodebailar:
refactor header area in overview fragment with better formated values for basal and iob


MilosKozak:
we should do this for nsclint layouts at once especialy when you stop using some Views

PR #2314, opened 2019-12-28
Watchfaces : Chart and interface improvements (colors and texts size) and customized direct Tap actions

Philoul:
Oups, cumulative PR (also include previous PR with Watch Tap Action)

This PR is to improve Colors of AAPSv2 watch face and also improve visibility on High res watch (my Ticwatch pro has 400x400px resolution)
Colors are standaard with AAPS App on phone :
=> BG point on graph are green if in range (white previously)
=> Carb point are Orange (green previously)
BG Value and arrow are green if in range
BG value, Hour and date are bold
Text size is proportional to resolution (very close to original if 320x320px watch)

See below differences : 
AAPSv2 Improve visibility on high res watches

I could not test on watches with a lower resolution, but the textsize should be very close to the original with a resolution of 320x320

Commits from 12/15/2019 to 12/18/2019 are for 5 customized zones for direct actions from watchface... See https://github.com/MilosKozak/AndroidAPS/pull/2284

### DIFF
--- a/app/google-services.json
+++ b/app/google-services.json
@@ -13,12 +13,7 @@
           "package_name": "info.nightscout.aapspumpcontrol"
         }
       },
-      "oauth_client": [
-        {
-          "client_id": "477603612366-a925drvlvs7qn7gt73r585erbqto8c79.apps.googleusercontent.com",
-          "client_type": 3
-        }
-      ],
+      "oauth_client": [],
       "api_key": [
         {
           "current_key": "AIzaSyDcZpDRMaGjdhihXp531cVYM6LkEL8KbgM"
@@ -42,12 +37,7 @@
           "package_name": "info.nightscout.androidaps"
         }
       },
-      "oauth_client": [
-        {
-          "client_id": "477603612366-a925drvlvs7qn7gt73r585erbqto8c79.apps.googleusercontent.com",
-          "client_type": 3
-        }
-      ],
+      "oauth_client": [],
       "api_key": [
         {
           "current_key": "AIzaSyDcZpDRMaGjdhihXp531cVYM6LkEL8KbgM"
@@ -71,12 +61,7 @@
           "package_name": "info.nightscout.nsclient"
         }
       },
-      "oauth_client": [
-        {
-          "client_id": "477603612366-a925drvlvs7qn7gt73r585erbqto8c79.apps.googleusercontent.com",
-          "client_type": 3
-        }
-      ],
+      "oauth_client": [],
       "api_key": [
         {
           "current_key": "AIzaSyDcZpDRMaGjdhihXp531cVYM6LkEL8KbgM"
@@ -100,12 +85,7 @@
           "package_name": "info.nightscout.nsclient2"
         }
       },
-      "oauth_client": [
-        {
-          "client_id": "477603612366-a925drvlvs7qn7gt73r585erbqto8c79.apps.googleusercontent.com",
-          "client_type": 3
-        }
-      ],
+      "oauth_client": [],
       "api_key": [
         {
           "current_key": "AIzaSyDcZpDRMaGjdhihXp531cVYM6LkEL8KbgM"
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -109,7 +109,7 @@ android {
         targetSdkVersion 28
         multiDexEnabled true
         versionCode 1500
-        version "2.6-dev"
+        version "2.6-rc2"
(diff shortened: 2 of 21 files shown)


jotomo:
Some semi-random notes, not a review:

>  => BG point on graph are green if in range (white previously)

White for in-range readings and current value was used on watch for better readability (in sunlight) over green

>  => Carb point are Orange (green previously)

I meant to use green for meals (combination of blue insulin dot and orange carbs dot, since displaying both separately would result in one overlapping the other). But it seems I dropped the ball midway through ...

> I could not test on watches with a lower resolution

You can try with an emulated watch - this helps making sure it didn't accidentally break anything basic for low-res watches.

Generally, naming conventions should be followed: field and method names start with a lowercase character in Java/Kotlin.

Philoul:
I reduce just a litlle bit text size for I think worst case (slow carbs + detailled IOB + date)...
screen 17

Philoul:
@jotomo Thanks for your comment,
on my side, I did not notice any visibility problem with green, but given the remark, I propose a lighter green if the standard green is difficult to see on other watches... (but we can go back to white if you think it's really better for all...)
I renamed the 2 methods so that they conform to the Java/Kotlin convention (doAction and doTapAction).
screen 21

For Bolus and Carb overlap, I don't think it's a problem if Bolus point is smaller than Carb point and if Bolus is in front of carb point... (For SMB it's the case and I think it's Ok.
For standard bolus, Currently Bolus point is about the same size than Carb and is behind carb point...
=> May be if we reduce just a little bit Standard Bolus point size and put it in front of the Carb, we will see both point and it'll be ok.

I tried to install Wear watch emulator for low res watches, but I don't know how to use it... (I can launch watchface, but I don't have data in it, I suppose I need to use "test" folder but don't know what to do...)
emulator

Philoul:
For bolus and Carb overlap, after different trials, I propose to keep the size of the bolus (so that we can easily differentiate a bolus with its greater size than an smb), I have put bolus point in front of the Carb point and I have increased the bolus down offset... Now we can easily see Bolus and Carb simultaneously. I think it's better than a green point...
And result on my Ticwatch pro : 
screen 23

app/google-services.json
```diff
@@ -13,12 +13,7 @@
           "package_name": "info.nightscout.aapspumpcontrol"
         }
       },
-      "oauth_client": [
-        {
-          "client_id": "477603612366-a925drvlvs7qn7gt73r585erbqto8c79.apps.googleusercontent.com",
-          "client_type": 3
-        }
-      ],
+      "oauth_client": [],
```
MilosKozak: it's not based on den as it removes some other code

Philoul: I don't understand "it's not based on den as it removes some other code" (do you mean "dev" or is "den" something else) ?

I think I must have mismanaged my branches.
I wanted to do 2 independent PRs, a "Watch Tap Action" Branch for direct actions from watchfaces and a "Color Improvement" branch for improving colors and the graphic interface of WF AAPSv2.
=> I wanted to start my 2 branches from the dev version, but I must have done something wrong because I noticed that the "Color Improvement" branch included (independently) the modifications of "Watch Tap Action" (on a separate PR)...
=> So with this PR, the PR "Watch Tap Action" becomes useless ... (I have not deleted it in case the PR "Color Improvement" is not accepted...)

What can I do to correct that ?

jotomo:
@Philoul 
> I tried to install Wear watch emulator for low res watches, but I don't know how to use it... (I can launch watchface, but I don't have data in it, I suppose I need to use "test" folder but don't know what to do...)

You need AAPS on a physical phone to provide data. Setup is described here: https://developer.android.com/training/wearables/apps/creating#pair-phone-with-avd

AdrianLxM:
What you could do is rebasing your work on current `dev`.
Then only the watch changes should go in. (Please use a new branch for that and don't force-push to an open PR)

Tanks :)

Philoul:
@AdrianLxM 
Thank's for your help, I have done rebase on a new branch called Watch_Improvement.
I think It's cleaner, when I compare Milos/dev and my new Watch_Improvement branch, I find only my commits (for both direct tap actions on watchfaces and color or texts size) except one commit done by Milos on build.gradle for main app (2.6-dev replaced by 2.6-rc2)...

Watch Commits

=> is that Ok or should I do something else (specially for build.gradle...) ?

Philoul:
Ok, I have done an interactive rebase and now I think it's clean...
I'll send a new PR with both Direct tap action on watchfaces (all), color and text size improvements

PR #2332, opened 2020-01-03
Fix for mixed up bage_warn, bage_urgent for Combo in Statuslights

swissalpine:
Fixes the small bug mentioned in https://github.com/MilosKozak/AndroidAPS/issues/2320

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/StatuslightHandler.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/StatuslightHandler.java
@@ -121,7 +121,7 @@ class StatuslightHandler {
     }
 
     void handleAge(String nsSettingPlugin, String eventName, TextView view, String text,
-                   int defaultUrgentThreshold, int defaultWarnThreshold) {
+                   int defaultWarnThreshold, int defaultUrgentThreshold) {
         NSSettingsStatus nsSettings = new NSSettingsStatus().getInstance();
 
         if (view != null) {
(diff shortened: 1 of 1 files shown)


gregorybel:
What are you fixing here?
Should not a warning value be smaller than an urgent value?
Why do you not simply change value in nightscout?

swissalpine:
The first value is warn, the second urgent warn  and it is mixed up.

swissalpine:
With this fix the warning value is smaller than the urgent warning value:
image

gregorybel:
On my side nightscout values work.
Ok now got it, in handleAge, warn and urgent are inverted:


    void handleAge(String nsSettingPlugin, String eventName, TextView view, String text, int defaultUrgentThreshold, int defaultWarnThreshold)

swissalpine:
Thanks for the hint, @gregorybel, so I changed handleAge() too and I adjusted the values to 244/336 as found in adjustAge().

gregorybel:
you are welcomed.
But should we not invert at all place using method handleAge?

swissalpine:
I'm not sure I understand your question correctly. In the Careportal the order is also warn - urgent. So now they are all in the same order, at least I think so ...

gregorybel:
I mean, the method is called several times and I wonder if we should not Swap the values at each call

swissalpine:
Many thanks for the explanation. This weekend I'm going to search if there are still functions with swapped values. At first sight I don't see any more.

MilosKozak:
is this finished?

swissalpine:
Yes, I think so. I haven't found any more.

PR #2333, opened 2020-01-03
fix: basebasal to standard color if no temp basal

osodebailar:
Basebasal was unvisible due to wrong  textcolor if basal was not a temp basal.


MilosKozak:
this is already fxed in dev

PR #2334, opened 2020-01-04
Watchface improvements : (Direct actions, color and text size)

Philoul:
Clean PR for Watchface improvements : 
- 5 customized zones for direct tap actions from watchfaces
- Color improvement in graph (Carb color orange, Bg value in range light green in dark watchface, dark green in bright watchface)
- Color for BG value in range (light green in dark watchface, dark green in bright watchface)
- AAPSv2 : Adjust text size according to watch resolution to improve lisibility
I use it on my Ticwatch pro and it's fine...

AAPSv2 watchface
screen 22

Settings for each zone
screen 25

for detailled history of commits see https://github.com/MilosKozak/AndroidAPS/pull/2314

I still have to check with different watch resolutions as @jotomo explained to me for text size (AAPSv2 only)

### DIFF
--- a/wear/src/main/java/info/nightscout/androidaps/watchfaces/BIGChart.java
+++ b/wear/src/main/java/info/nightscout/androidaps/watchfaces/BIGChart.java
@@ -41,12 +41,18 @@ import com.ustwo.clockwise.wearable.WatchFace;
 import java.util.ArrayList;
 
 import info.nightscout.androidaps.R;
+import info.nightscout.androidaps.aaps;
 import info.nightscout.androidaps.data.BasalWatchData;
 import info.nightscout.androidaps.data.BgWatchData;
 import info.nightscout.androidaps.data.BolusWatchData;
 import info.nightscout.androidaps.data.ListenerService;
 import info.nightscout.androidaps.data.TempWatchData;
+import info.nightscout.androidaps.interaction.actions.BolusActivity;
+import info.nightscout.androidaps.interaction.actions.ECarbActivity;
+import info.nightscout.androidaps.interaction.actions.TempTargetActivity;
+import info.nightscout.androidaps.interaction.actions.WizardActivity;
 import info.nightscout.androidaps.interaction.menus.MainMenuActivity;
+import info.nightscout.androidaps.interaction.menus.StatusMenuActivity;
 import lecho.lib.hellocharts.view.LineChartView;
 
 /**
@@ -67,7 +73,7 @@ public class BIGChart extends WatchFace implements SharedPreferences.OnSharedPre
     public int basalBackgroundColor = Color.BLUE;
     public int basalCenterColor = Color.BLUE;
     public int bolusColor = Color.MAGENTA;
-    public int carbsColor = Color.GREEN;
+    public int carbsColor = Color.rgb(255,160,0);
     public int pointSize = 2;
     public boolean lowResMode = false;
     public boolean layoutSet = false;
@@ -98,8 +104,8 @@ public class BIGChart extends WatchFace implements SharedPreferences.OnSharedPre
     private String cobString = "";
 
     private TextView statusView;
-    private long chartTapTime = 0l;
-    private long sgvTapTime = 0l;
+    private long TapTime = 0;
+    private WatchfaceZone LastZone = WatchfaceZone.NONE;
 
     @Override
     public void onCreate() {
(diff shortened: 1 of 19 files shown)


Philoul:
@jotomo 

> You need AAPS on a physical phone to provide data. Setup is described here: https://developer.android.com/training/wearables/apps/creating#pair-phone-with-avd

This week end I tried  to use emulator for low res watches, but It doesn't work, I can't provide data with my 2 phones...
=> My main smartphone (with Android APS, and Wear OS connected on my Ticwatch Pro) is professional and locked by my company (impossible to active developer mode on this phone locked by Knox, and I need to use a workaround to install AAPS...)
=> My second phone, used for development (with AAPS 2.6RC2 or dev installed on it) is a low cost phone running with Android Go... I can active developer mode on it but Wear OS app is not compatible with this phone, so impossible to install Wear OS app on it :-(
(I tried to manually install Wear OS app with an apk but app freeze...) 
So I can't pair emulator on my 2 phones to send data...

Here is calculation done for text size adjustment (only AAPSv2 watchface) : About the same for 320px resolution, A little bit smaller for 280px or lower resolution watchs...
`// Adjust text size according to watchscreen resolution`
`hoursize = mRelativeLayout.getHeight()/9.5;   // 42 for 400, 34 for 320 , 29 for 280 (original = 30)`
`svgsize = mRelativeLayout.getHeight()/8;       // 50 for 400, 40 for 320 , 35 for 280 (original = 38)`
`smalltxt = mRelativeLayout.getHeight()/32;    // 13 for 400, 10 for 320 , 9 for 280 (original = 10)`
`midtxt = mRelativeLayout.getHeight()/25;      // 16 for 400, 13 for 320 , 11 for 280 (original = 14)`

Do you think it's ok ? (If you prefer I can also add the original text size for Hour, BG, IOB and COB as minimum values ​​... This way, the text cannot be smaller than the current size for low resolution watches)

jotomo:
@Philoul Travis build fails. Seems like the file _WatchfaceZone_ only exists on your system and hasn't been committed: 
https://travis-ci.org/MilosKozak/AndroidAPS/builds/632505063#L3337

Philoul:
Thanks @jotomo , 2 forgotten files added in last commit

dlvoy:
Hi, my first remarks after tests:
* bold & bigger SGV & clock make layout issues (see SCR below)
* down direct action conflicts with double click on graph
* sometimes click failed
* there is significant delay (2s ... 5s) between click action and result (configured dialog is shown) - it is due nature of Wear OS where opening any activity can be sluggish, but without immediate (<200ms) feedback after click as a user i am not sure, that click is working. There should be some hint (visual or vibration) that click was made to inform that action trigger worked and action will execute.
* for action configuration i would prefer one item in setting that opens live preview - custom dialog that shows action areas as WYSIWYG parts of screen, clicking which configures them, otherwise i do not know area location and what i configure, where is "top" and where "right" start? - I've seen that by other commercial watchfaces

device-2020-01-22-100150
device-2020-01-22-100329


**my watch:** Huawei Watch 2 0813 (LEO-DLXX), Wear OS 2.12 / Android 8.0.0

Philoul:
Hello @dlvoy 
Thank you for your feedback. I think I will close this PR and split it into 3:
- One for color improvement
- One for double tap actions
- One for text size

=> For layout problems, our watch resolutions are very close (mine is 400x400 and yours 390x390) and I am surprised that there is such an important difference... (If i compare screenshots in photoshop, all Text size in your watch are bigger in pixels than in mine, even with a smaller screen resolution and a smaller value for `xxx.setTextSize` setting...)
In my code, the TextSize setting is proportional to the resolution of the watch: i.e. for hour 
`hoursize = (int)(mRelativeLayout.getHeight()/9.5)`
     My watch (400px)   : `mTime.setTextSize(42)`, and result in pixel = ~52px
     Your watch (390px) : `mTime.setTextSize(41)`, and result in pixel = ~59px
Do you know how I can calculate textsize setting to be sure that results in pixel is proportional to watch resolution ?

=> For direct action, I agree that the delay can be quite long, and having immediate feedback is a good idea (vibration is the easiest way for me I think), and that feedback could also be added in current watchfaces for main menu...

=> For setting, I think it should be organized (including a graphical WYSIWYG probably better for Direct actions settings), but I think it is too complicated for me today: I tried to include `<PreferenceScreen` sections in preferences.xml file  to add several submenus, but I need to remove finish() line in AAPSPreferences.java file to allow navigation in submenus, but I found side effects and I failed to keep AAPS background in submenus...

dlvoy:
@Philoul as a reference here how this watchface looked before changes on my watch,
this is from *2.6-RC5*:
device-2020-01-22-105132

* Splitting it into separate PRs is good idea :+1: - easier to test and review, one PR per feature

* For font sizes / and layout - it is always tricky as fonts and dpi-s can vary a lot, so **assume nothing**. Preferably it should be set in layout, not in code - but sometimes you have no choice, and then:
  * for text: `textView.setTextSize(TypedValue.COMPLEX_UNIT_SP, 65);`
    `SP` if you want text size be OS font size dependent, it will be density aware but change if user enlarge texts size on phone, if only density dependency is needed, use `DIP` instead https://developer.android.com/reference/android/util/TypedValue
  *  for other layout calculations:
      ```java
      DisplayMetrics dm = new DisplayMetrics();
      getWindowManager().getDefaultDisplay().getMetrics(dm);
      pixelSize = (int)scaledPixelSize * dm.scaledDensity; 
      ```
* For click haptic feedback i've usually seen background hilite (on buttons and complications), but since we have zones here it will be more confusing and harder to implement, so vibrations :+1: 

* I forgot to mention that annoying finish after every change in settings :rage: Not sure why it is needed at all - my gut feeling is because of simplistic/workaround way watchfaces and menu read prefs in the first place, that force kill-to-refresh-prefs. :man_facepalming: There is better way -  event for prefs change that should be handled etc.
So maybe it is easier to keep it simple right now and leave prefs as it is - and improve on them after refactoring - i volunteer for that :muscle: But to avoid conflicts, lets wait till above changes get done, reviewed and merged :wink:

Philoul:
@dlvoy I checked font size in my watch settings (medium) and if I change this setting to "Large font", I have same issue than yours... Do you have "Large" Font in your watch?
I also found issues in some current watchfaces with my watch, probably linked to font size watch setting...
We have lots of informations in our watchfaces so it may be better to have text sizes independent of the system setting of the watch... Do you think it's a good idea?

dlvoy:
@Philoul i have "Normal" (there is also small and large to choose from)
It is both OS / Rom and user config dependent. There are two approaches:
* use it, allowing users with impaired vision etc. to enlarge at will - but it is harder to make layout that always work
* make it only DPI dependent - easier to do and test
Probably - for now - 2nd option is more feasible

Philoul:
I close this PR (split in 3...) First one only for colors done https://github.com/MilosKozak/AndroidAPS/pull/2390

PR #2338, opened 2020-01-04
Watch loop action

zjedr:
Add ability to disconnect/suspend/disable, resume/enable loop from a new menu option on watch. On the AAPSv2 watchface the divider goes red when not looping.

Loop1
Loop2
Loop3
Loop4
Loop5
Loop6
Loop7
Loop8
Loop9


zjedr:
closed

PR #2342, opened 2020-01-08
Watch loop action

zjedr:
Add ability to disconnect/suspend/disable, resume/enable loop from a new menu option on watch. On the AAPSv2 watchface the divider goes red when not looping.

Loop1
Loop2
Loop3
Loop4
Loop5
Loop6
Loop7
Loop8
Loop9

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
@@ -686,6 +686,7 @@ public class LoopPlugin extends PluginBase {
             });
         }
         NSUpload.uploadOpenAPSOffline(durationInMinutes);
+        RxBus.INSTANCE.send(new EventLoopUpdateGui());
     }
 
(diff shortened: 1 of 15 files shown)


Philoul:
I saw you add strings in this PR compared to your previous one 👍 
Why did you add `<string name="duration">Duration</string>` in string file ?
There is allready one string used in same context in strings.xml file : `<string name="action_duration">duration</string>`

On my side, I made a PR https://github.com/MilosKozak/AndroidAPS/pull/2334 to add "Direct Actions" from the 8 watchfaces with 5 custom zones... (some impacted files are the same, especially Home2.java)
=> I think there is a risk of conflicts between our 2 PRs

There is another setting ("match_divider") where status bar divider background matches watchface background... If you set it On, the divider bar is black and text is white... Did you check if contrast between your Red divider and white text is enought when loop is suspended or disconnected ?

zjedr:
match_divider looks ok to me, I can swap to black text if that is the consensus.
LoopA

I'll missed the existing string, I'll swap to it on the next commit.

PR #2366, opened 2020-01-15
IOB fix for tiny basals on Dana pumps

Mackwe:
Any basal rate less than 0.10u/h will effectively become a zero-temp until minute :56 of the hour, where the entire hour's basal (0.01-0.09u) is released. This is by design in the Dana pumps and cannot be altered by an APS algorithm. However, the APS _thinks_ the basal is delivered every 5 mins, so it may create IOB that isn't there.

This patch makes the pumps go for zero temping unless 0.10u/h or more is requested. This way the IOB will match during nighttime (where this is likely to be a problem for small kids).


MilosKozak:
do you have this threshold or behavior confirmed by sooil?

Mackwe:
Yes. I do. I believe it is also in the manual.

Den ons 15 jan. 2020 19:41Milos Kozak <notifications@github.com> skrev:

Mackwe:
Den ons 15 jan. 2020 19:41Milos Kozak <notifications@github.com> skrev:

Mackwe:
Screenshot_20200115-194525_OneDrive

guydavies:
It doesn't appear to say that it's delivered in minute 59 but that there's a period of delivery of 60 minutes. That could still be true as long as the insulin is delivered at exactly the same instant in the 60 minute period.

Mackwe:
I have this verified by Sooil. Basals < 0.10u/h are delivered @ minute 59.

guydavies:
OK. That seems like the conservative approach. Don't give the insulin until the end of the delivery period. But it does have some interesting potential consequences for me since I have a fair portion of the afternoon below 0.1u/h.

Mackwe:
> it does have some interesting potential consequences for me since I have a fair portion of the afternoon below 0.1u/h.

Yes. And our son has multiple hours at night with basals below 0.10u/h. And when the algorithm makes a low TB it can become less than 0.10u/h even if the base basal is way above.. Which just makes IOB. No insulin given.

MilosKozak:
@Mackwe 
if we apply this patch
what if we set zero temp for 0-58 min and 0.1 for last 58-60min?

guydavies:
Hi Milos

Would that be right?  If you set 0.1u/h from 59-60, that would only be 1/60
of 0.1. You'd need to go for 6u/h for one minute (or actually deliver a
"bolus" of 0.1).

Rgds

Guy

Mackwe:
> what if we set zero temp for 0-58 min and 0.1 for last 58-60min?

The way I have created the PR, it will resort to "if total momentary basal requirement is less than .10u/h, set a zero temp". This also means: "whenever the basal need is .10u/h or higher, set a temp to meet the need". If you have a base basal need of 0.09u/h, you will likely get a 0.10u/h just after one hour passes. The base basal keep accumulation until the .1u/h threshold is met. You can of course implement your fix as well, resulting in ~0,007u being delivered. I would however use 0.15u/h in that case, delivering precisely 0.01u. That's still 1/4 of the lowest basal possible.

> If you set 0.1u/h from 59-60, that would only be 1/60 of 0.1. You'd need to go for 6u/h for one minute (or actually deliver a "bolus" of 0.1).

I guess what Milos meant was to deliver the smallest amount possible, not 0.10u, as this may be way too much for small kids. And to deliver 0.10u you would have to set a 0.10u * 15 = 1,5u/h basal as the pump delivers 15 times an hour (every 4 mins).

guydavies:
Thanks for the clarification, Marcus.  That makes sense.

Issue #2357, opened 2020-01-12
[2.6 RC3/4] Bg readings view cut


pmisun:
![20200112_100026](https://user-images.githubusercontent.com/12038874/72216529-749fa080-3522-11ea-958f-0dd642a11024.jpg)

![20200112_114713](https://user-images.githubusercontent.com/12038874/72217665-8472b100-3531-11ea-97f5-bf1b1366c93e.jpg)

Phone Samsung xcover4s
720 x 1280 px / 294 ppi

pmisun:
![20200113_195814](https://user-images.githubusercontent.com/12038874/72283482-26cc8a80-363f-11ea-801d-23b48e093400.jpg)


Will add here some other small ui glitches

dlvoy:
Both confirmed, by SGV, when not-yet connected / received current data from xDrip it cut the SGV section

![sgv_hidden](https://user-images.githubusercontent.com/5094588/72330024-199eb280-36b6-11ea-92c6-65eb46d51a2e.jpg)

For insulin delivery progress dialog, this nice title is not fitting when text is too long in translation (like in my PL case):
![delivery_title](https://user-images.githubusercontent.com/5094588/72330159-4652ca00-36b6-11ea-8618-c7a367ae7bd4.jpg)

What is approach here for this kind of bugs - since it can be fixed with font auto-scaling (with: `android:autoSizeTextType="uniform"`,`android:autoSizeMinTextSize="12sp"` ... etc) like [this layout fix for Wear OS](https://github.com/MilosKozak/AndroidAPS/blob/2.6-rc3/wear/src/main/res/layout-v26/list_item.xml)
but that will work only API 26+ so require layout-v26

pmisun:
This is frim RC4. Making the graph too small and the cut of arrow still appears. 

![Screenshot_20200114-133920_AndroidAPS](https://user-images.githubusercontent.com/12038874/72347135-b4f44f80-36d7-11ea-829e-3c80ffde18fb.jpg)

swissalpine:
Could / should we roll back to the "old" design. I see no benefits and it produces layout problems.

MilosKozak:
i fixed it in RC5

MilosKozak:
bolus progress is still open

PR #2380, opened 2020-01-19
[2.6-rc5][fix] Layout fix for bolus progress dialog header text

dlvoy:
This fix is for pump progress dialog (examples of bug in #2357) - for longer texts (like in some translation) header text was too long and floated over (obscured by) header icon - fixed by making text to the right of icon in layout

**Was:**

Screenshot_20200119-071114_AndroidAPS

**After fix:**

Screenshot_20200119-075012_AndroidAPS

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -109,7 +109,7 @@ android {
         targetSdkVersion 28
         multiDexEnabled true
         versionCode 1500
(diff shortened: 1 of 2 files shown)


MilosKozak:
i created new commit istead of resolving this.
thanks

PR #2390, opened 2020-01-25
Watchfaces color improvements

Philoul:
Separate PR for Watch Improvements
This first PR contains only Colors Improvements:
- Carbs are Orange in chart
- BG, arrow and MidColor changed from white (or black) to light green or dark green (standard green not used for better visibility in sun light)

First commit for colors.xml
One commit per Watchface (commit for AAPSv2 also include modification of BaseWatchFace file)
No change for Cockpit, NoChart and Circle Watchfaces

Result on AAPSv2 watchface
screen 34
screen 35
Note: date seems to be grey but colors are Black or White, but font is too small and too thin in current watchface


dlvoy:
Tested on Huawei Watch 2 - works fine. 
From code point review - it is ok. I would like (and will do in future) big refactoring of watchfaces regarding how colors (themes) are set up and propagated to UI. But this is story for another PR :)

device-2020-01-25-190500
device-2020-01-25-190704
device-2020-01-25-191113
device-2020-01-25-191120

PR #2391, opened 2020-01-25
Some labels updated in Wear Manifest to allow translations

Philoul:
Today these labels are not translated in confirmation's popup of actions


MilosKozak:
these strings don't exist

PR #2421, opened 2020-01-31
Restore ns_noupload function

twain47:
The ns_noupload is broken in 2.6. Looks like it was lost as part of the refactor of DBAccessReciever.

There may be better places to implement this but this seems to work.


MilosKozak:
good catch and correct solution.
It has conflict with dev so i will close this PR but add similiar code.
Thanks

PR #2454, opened 2020-02-20
Use google key/value backup agent for settings and objectives

twain47:
This switches from auto backup to key-value backup.

Auto backup doesn't seem to be a good choice for AAPS backup because:

1) Auto backups only run if the app is not active (which almost never happens with AAPS)
2) AAPS stores a large amount of data which tends to exceed the 25MB limit imposed causing backups to be disabled

This PR avoids the issues by only backing up shared preferences and using the older key-value backup system which doesn't require the app to be shutdown.

On re-installing AAPS the wizard is not run but the user is correctly prompted for permissions at the top of the screen. It seems to sometimes be necessary to restart AAPS for the settings to be fully applied.

This isn't as reliable as the import/export settings, but it gives another chance to recover if an export is not available.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -37,7 +37,14 @@
         android:roundIcon="${appIconRound}"
         android:supportsRtl="true"
         android:theme="@style/AppTheme.Launcher"
-        android:fullBackupContent="true">
+        android:fullBackupOnly="false"
+        android:backupAgent=".utils.SPBackupAgent"
+        android:restoreAnyVersion="true">
+
+        <meta-data
+            android:name="com.google.android.backup.api_key"
+            android:value="AEdPqrEAAAAI3JiApyMrbP2QFzZ2fYfCPsgjkRp53Dm2S1-zPQ" />
+
         <meta-data
             android:name="com.google.android.gms.car.application"
             android:resource="@xml/automotive_app_desc" />
--- a/app/src/main/java/info/nightscout/androidaps/utils/SPBackupAgent.java
+++ b/app/src/main/java/info/nightscout/androidaps/utils/SPBackupAgent.java
@@ -1,0 +1,24 @@
+package info.nightscout.androidaps.utils;
+
+import android.app.backup.BackupAgentHelper;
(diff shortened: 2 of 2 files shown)


MilosKozak:
how to restore this backup?

twain47:
Android appears to do it automatically when the apk is installed.  The docs are not the best though so it isn't well defined.  I've not found any way to trigger one from within the app.

You can also do a manual backup / restore. Documentation is here: https://developer.android.com/guide/topics/data/testingbackup

`adb shell bmgr backupnow --monitor-verbose info.nightscout.androidaps`

`adb shell bmgr restore 3ce7c18a7ba51e65 info.nightscout.androidaps`
(where the hex token is from `adb shell dumpsys backup | grep Current:`

app/src/main/AndroidManifest.xml
```diff
@@ -37,7 +37,14 @@
...
         android:supportsRtl="true"
         android:theme="@style/AppTheme.Launcher"
-        android:fullBackupContent="true">
+        android:fullBackupOnly="false"
+        android:backupAgent=".utils.SPBackupAgent"
+        android:restoreAnyVersion="true">
+
+        <meta-data
+            android:name="com.google.android.backup.api_key"
+            android:value="AEdPqrEAAAAI3JiApyMrbP2QFzZ2fYfCPsgjkRp53Dm2S1-zPQ" />
```
MilosKozak: where is this api key from?

MilosKozak:
is it posssible to ask user before restore?

app/src/main/AndroidManifest.xml
twain47: https://developer.android.com/google/backup/signup

As far as I can see there are no special rights (I can't see anyones backup data, in fact even the user can't access their own backup data except to delete it). Above link will let you generate a new one.

twain47:
> is it possible to ask user before restore?

Not using googles standard code.  Googles idea is that this restores the state as last installed (usually when moving to a new device).

This is really just a fix for the existing 'android:fullBackupContent="true"' that you added but doesn't quite work.

What are you trying to fix?  It might help me suggest ways round the problem.

PR #2459, opened 2020-02-25
[WIP] Allow combo pump and phone to use different timezones

twain47:
By storing a timezone for the pump the plugin is able to apply a time offset during communications with the pump and avoid the sync issues experienced when moving between timezones or DST.  

Basal profiles are offset automatically and re-applied to the pump so if functions correctly if it loses connection with AAPS.

The main disadvantage is that the pump clock is 'wrong' so any data pulled directly from the pump by looking at the screen or by a clinic will show the wrong time, however I wouldn't think that would be a significant problem for most AAPS users.

Defaults to using zero offset (phone and pump time are synced) and must be manually activated.

Looking for feedback - there may be issues that I've missed.


jotomo:
@twain47 This adds non-trivial complexity and breaks some core assumptions ... (mainly the interpretation of the partial date shown on the display, which requires some thought around day changes and year changes as well as warnings/alarms).
The sync issues you're referring are having to manually set the time on the pump? 
And/or IOB calculations when doing so?
I'll assume travel across timezones is the main issue (rather than having the pump's clock off half the time due to skipping DST).
So the use-case is this then(?): a user travels and the phone (auto)updates the time. The user then sets an offset and AAPS begins shifting the basal profile by that offset.
It might be because I'm not time-zone travelling, but I'm not really seeing how this PR simplifies things for the user. Could you elaborate?

On a design level: as mentioned above, assumptions need to be rechecked - everywhere where time is involved. I assume you haven't tested the code yet? It should raise rather annoying alarms about the pump's time being different from the phone (_ComboPump.checkPumpTime()_). 

Code-wise, I'd like a few things changed, the code has to much state as it is, so any further state should be handled carefully (passing all arguments to command classes via ctor, extracting time/offset manipulations into methods, naming) or avoided if possible. But let's discuss the use case before diving into details.

As an alternate path: one thing that isn't yet implemented but is basically the last (useful) possible feature to add, is to extend ruffy to be able to handle the _set date/time_ menu, so AAPS could update the pump clock. It's not a 2h job though, otherwise it would have been done already; though that would be a cleaner way to solve this, which I'd prefer.

twain47:
@jotomo Thanks for looking at this.

I'll stick to commenting mostly on the use-case elements since they seem the main focus.

For us the use-case would be both timezone and DST. Personally I'd prefer to leave the pump permanently on a single timezone and avoid having to change the time again, but probably it is most useful for frequent travellers who have to go through the timezone process regularly.  

The main issue I've seen (and maybe there is another workaround that I've missed?) is that changing the pump clock backwards results in bolus history in the future. This either means clearing future records or disabling looping until the history elements are past.  Here is a link to the process for combo users: https://androidaps.readthedocs.io/en/latest/EN/Usage/Timezone-traveling.html?#accu-chek-combo

If the pump timezone is set before travel / DST changes then it can be ignored - everything just continues to work, all be it with the pump still on the old time setting. The timezone change notification causes AAPS to push an offset basal pattern in case of loss of connection.

I agree on the level of extra complexity.  I've done my best to test this, although my testing has been on my main dev branch of which this is an extract. I'm in the fortunately position of having a spare pump & phone so this has been running on that as a slave for the last week and I've repeatedly moved the timezone of the phone during this period. All appears to be working fine.  Since all timestamps from the pump are offset checkPumpTime sees the pump as having the correct time and no error is generated.

Re: automating setting time via ruffy - I was considering this a next step. It would be nice to be able to sync the time on the pump to avoid clock drift regardless of this PR.  Simply updating the time on the pump causes the history problems previously mentioned so isn't a suitable solution by itself.

th122:
To add more confusion to the issue: My preference would be to set the pump clock to UTC, tell the loop about this :), and forget all about having to worry about DST, travelling, etc.
For the process reading out the pump, timestamps would always be linear, without inconsistencies in time  during continuous operation.
The Loop could rely on its calculation of IOB, and display the continuous data within the currently selected local timezone...

twain47:
> To add more confusion to the issue: My preference would be to set the pump clock to UTC, tell the loop about this :), and forget all about having to worry about DST, travelling, etc.

@th122 
This patch allows you to do this. Just specify the pump as timezone +0:00 (which is UTC)

app/src/main/res/xml/pref_combo.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <PreferenceCategory android:title="@string/combopump">
+
+        <SwitchPreference
+            android:defaultValue="false"
+            android:key="@string/key_pump_uses_utc"
+            android:title="@string/pumpt_uses_utc" />
```
jotomo, line 9: Switching the pump to UTC will create duplicates (for the hours shifted from local time to UTC).
It would be helpful to add a _android:description_ element here that asks the user to (refresh the pump and) check treatments when switching pump to/from UTC.

jotomo: Sorry, that should be _android:summary_.

Issue #2472, opened 2020-03-02
SMS carbs functionality not working


glaskopf:
**I have  a problem with the new command "carbs"** 

- All other SMScommands are working for me. 
- The problem is reproduceable for me

**Description**

- Sending SMS:"_carbs 6_". 
- Receiving SMS: "_Um 6g Kohlenhydrate einzugeben, antworte mit dem Code Esa_"
- Sending SMS:"_Esa_"
- REceiving SMS: "_6g Kohlenhydrate erfolgreich erfasst_"

So far, so bad because AndroidAPS on the smartphone shows in parallel  an alarm and the message on smartphone:

"_Der Eintrag (Insulin 0,00, Kohlenhydrate:6 ) konnte nicht als Behandlung gespeichert werden. Bitte überprüfe die aktuelle Liste und füge, falls notwendig, den Datensatz manuell hinzu_"

And the error message is correct, because the carbs entry was not considered /saved.

---
AndroidAPS 2.6.0

AdrianLxM:
Could you please provide log files?

glaskopf:
Yesm I sent it via eMail. Please ignore the first email, because the problem was not in file.
In the newer logfile you can find the problem at ca. 09:57:25.464

MilosKozak:
[AndroidAPS.log](https://github.com/MilosKozak/AndroidAPS/files/4289927/AndroidAPS.log)
09:57:25.465

@jotomo || @AdrianLxM 
can you test it on combo? It works with Dana

glaskopf:
I have still the (exactly) same problem also with 2.6.1

MilosKozak:
can you test https://github.com/MilosKozak/AndroidAPS/pull/2528 ?

MilosKozak:
@glaskopf

glaskopf:
@MilosKozak 
What is the best way to test it? 
I can Build a new app release via AS -> but I'm not sure which branch I should use?
Or is there a more elegant way to build a new release only with the four changed files?

jotomo:
@glaskopf IMO the simplest way is to follow the steps in the first part of this blog post: https://blog.jetbrains.com/idea/2018/10/intellij-idea-2018-3-eap-github-pull-requests-and-more/ (Android Studio is based on IntelliJ, so the blog post also applies to Android Studio)
Then you have a local copy of the branch the PR uses, which you can then build and install as you usually do.

MilosKozak:
[...]

PR #2482, opened 2020-03-04
Combo: like with bolus record, allow updating existing record for car…

jotomo:
…b-only treatment as well.

For carb-only treatments Combo-code simply does nothing beyond calling _addToHistoryTreatment_. Align the code to allown merging of records in this case as well. Such a change was done previously for a request to add bolus (and carbs).

Untested as of now

#2472 @AdrianLxM @MilosKozak


MilosKozak:
looking at the. setting true has only impact on displaying error message. no impact on db at all

jotomo:
@MilosKozak You're right, then the issue is somewhere else ...

PR #2483, opened 2020-03-06
Combo: create separate bolus/carb records at driver level.

jotomo:
Tests pending, will follow shortly


jotomo:
@MilosKozak Tested now, works on my end

jotomo:
@swissalpine Please verify you tested the version I pushed an hour ago, before marking the PR ready for review (commit hash 70053ea)

jotomo:
Confirmation via another channel from @swissalpine that the correct version was tested).

PR #2493, opened 2020-03-09
Use "Display time-sensitive notifications" for Insight alerts

### LINKED ISSUE #2441: Issues with Insight Pump Alerts
When a pump alert is signaled, it will be repeated after some time in AndroidAPS.
In everyday life this is mostly used for 'Empty Insulin' messages.
This first starts some Ringtone-Sound and vibration and eventually displays a Activity, to confirm the alarm on the pump.

Having the possibility to confirm the alarm on Screen is great! 
But there are several issues:
- A Ringtone for a phonecall is used, which is a no-go
- The alarm sounds first, but the Dialog/Activity for confirmation sometimes pops up very slowly
  does not appear at all. (Especially if AAPS is busy)
  This means that you have a very loud phone (e.g. in school), but no possibility to confirm and silence it.
- The alarm activity is apparently started multiple times: I observed a stack of at least 3 of them, when I closed them with the BACK button.

Suggestions: 
- There is a race for launching the activity. This is scheduled in a handler thread, but repeated every second (in alarm polling) until the started activity has registered with the InsightAlarmService.
  Only the last activity will work properly.
- The alarm sounds could be left out completely. The pump has a nice scheme of increasing vibrations and sounds. The phone could sufficiently function as 'confirmation device'.
- If using sound is important, it should be no ringtone and it should use some increasing volume scheme so that in the best case the alarm can be confirmed before It annoys the crowd around you.
It must be granted that the Alert activity is displaying correctly, before the sound goes off.

TebbeUbben:
This uses "[display time-sensitive notifications](https://developer.android.com/training/notify-user/time-sensitive)", a combination of notification and activity, for the Insight alerts to fix #2441.

Screenshot_20200309-230940_Trebuchet

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -23,6 +23,7 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="com.dexcom.cgm.EXTERNAL_PERMISSION" />
+    <uses-permission android:name="android.permission.USE_FULL_SCREEN_INTENT" />
 
     <uses-permission-sdk-23 android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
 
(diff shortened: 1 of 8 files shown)


StephenBrown2:
Could this be generalized to other pumps?

fra-orolo:
Great! From looking at the code, this solves a couple of issues.

Issue #2505, opened 2020-03-14
Missing alert when connection to pump is lost... 


CaroGo:
Using Dana R and aaps build bbe2fd506-2020-03-09
My Batterie went dead without the typical sound.. Just stopped working in the night at about 23.50 o'clock. There was no alerting sound from aaps that pump connection was lost... I realised that about 2 or 3 hours later... 
I have local alarms set, and set 30 min for the alert for lost connection... 
Is this a problem within my settings which I can solve by myself or is it an issue which can affect others too? 
See log files... 
https://drive.google.com/file/d/1dlqbVZ16yoeqVvltTcRc8NIj8RDt8nN1/view?usp=drivesdk

https://drive.google.com/file/d/1dlqbVZ16yoeqVvltTcRc8NIj8RDt8nN1/view?usp=drivesdk

nehasvit:
It happened also to me today.
0:01 pump battery died, no pump alert. I also had sensor errors all night, so I had no high BG alerts.
AAPS unreachable alert was set also to 30 minutes, but alert didn't raise.
![photo_2020-03-14_22-46-06](https://user-images.githubusercontent.com/11347007/76690956-b396c300-6645-11ea-93a5-437b6ca982c8.jpg)

rkresha:
I've had similar problems for the last couple months. I dug in a couple weeks ago and have it working again, @MilosKozak and the core group will need to spot check my adjustments to make sure it doesn't have downstream impacts. My issues where a combination of not being rileylink connection issues and pump connection issues. PR will be coming soon

MilosKozak:
fixed by https://github.com/MilosKozak/AndroidAPS/commit/d9e08530fddc8cb9e07bcb3fd115ccb705aa8d94
thanks for reporting

PR #2508, opened 2020-03-15
Alert for pump unreachable not working

rkresha:
This has worked for me to get the alert to always work, where before it would more often not work and I'd find the last connection over 4 hrs ago and no alert.

Others having issues like seen in #2505


MilosKozak:
this is correct solution
https://github.com/MilosKozak/AndroidAPS/commit/d9e08530fddc8cb9e07bcb3fd115ccb705aa8d94
happend while kotlin conversion

PR #2528, opened 2020-03-22
SMS Carb things

jotomo:
For discussion/analysis

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
@@ -644,6 +644,7 @@ public class LocalInsightPlugin extends PluginBase implements PumpInterface, Con
                 result.comment = ExceptionTranslator.getString(e);
             }
         } else if (detailedBolusInfo.carbs > 0) {
+            TreatmentsPlugin.getPlugin().addToHistoryTreatment(detailedBolusInfo, true);
             result.success = true;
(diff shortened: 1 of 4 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -644,6 +644,7 @@ public PumpEnactResult deliverTreatment(DetailedBolusInfo detailedBolusInfo) {
                 result.comment = ExceptionTranslator.getString(e);
             }
         } else if (detailedBolusInfo.carbs > 0) {
+            TreatmentsPlugin.getPlugin().addToHistoryTreatment(detailedBolusInfo, true);
```
TebbeUbben: You didn't set a source here, so it will probably fail.

jotomo: I think a source is set in the passed DBI in all cases now (SMS Plugin was the offender who didn't do it so far).
If it's still missing, TreatmentService will at least log the error.
... this whole construct makes it way to easy to make errors though; not sure exactly how, but during 2.7 I'd like rework this. Maybe split DBI up similar to separate carb and bolus records in DB, so that there's a bolus command and a carbs command (CommandQueue). Then only the DanaR drivers needs to concern itself with merging carbs and bolus - if those entries are even stored in a single entry on the pump.

jotomo:
Superseeded by #2533

PR #2529, opened 2020-03-23
Fix order of insight histories

### LINKED ISSUE #2435: Paused pump - basal graph does not show zero temp
**Version**: 2.6RC7
**Issue**: when setting up a pump disconnect, the grayed out rectangle area is shown in the graph, but the displayed given basal rate is not degraded to zero in the graph. It looks more as a just normal running basal rate, but it should be on the bottom line accordingly to 0% that is correctly shown to the value „BAS“

### LINKED ISSUE #2484: [2.6.0] TBR sometimes not shown correct in start screen Insight pump
Sometimes AAPS dont show in the start screen the correct TBR (under BAS) like there, but the TBR in the pump ist set: 
buggedScreenshot_20200306-130315_AndroidAPS
buggedScreenshot_20200306-130330_AndroidAPS
sometimes it is correct (i mean the blue font)
okScreenshot_20200306-141344_AndroidAPS

Running on AAPS 2.6.0 on Galaxy S10+ with Insight and Libre 2
Logs (Date:06.03.2020, Time: see in screenshots): 
[AndroidAPS.log](https://github.com/MilosKozak/AndroidAPS/files/4298574/AndroidAPS.log)
[Log Export.zip](https://github.com/MilosKozak/AndroidAPS/files/4298588/Log.Export.zip)

Suchiman:
The insight pump only supports second precision so when a TBR is changed, this is a quick succession of "cancel TBR" + "set TBR" in the same second. The code sorts the events multiple times and processes the events from new to old for some reason. This leads to the "set TBR" record to be inserted first sometimes and then the database will update the "set TBR" record to be the "cancel TBR" record since they happened in the same second so we can't rely on timestamp to tell us if one event happened before another. Luckily the insight has a monotonically increasing eventPosition id we can sort by to have the events in order of happening. By just avoiding sorting all over the place we can keep and process the events from old to new. We still override the "cancel TBR" with the "set TBR" with this fix but this does not appear to be a problem.

Additionally, i've noticed that the historyEvents curiously enough had random milliseconds attached to them which were a artifact of how the timestamp were computed.

Fixes #2484
Fixes #2435

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
@@ -361,6 +361,7 @@ public class LocalInsightPlugin extends PluginBase implements PumpInterface, Con
         calendar.set(Calendar.HOUR_OF_DAY, pumpTime.getHour());
         calendar.set(Calendar.MINUTE, pumpTime.getMinute());
         calendar.set(Calendar.SECOND, pumpTime.getSecond());
+        calendar.set(Calendar.MILLISECOND, 0); // truncate leftover milliseconds from current time
         if (calendar.get(Calendar.HOUR_OF_DAY) != pumpTime.getHour() || Math.abs(calendar.getTimeInMillis() - System.currentTimeMillis()) > 10000) {
             calendar.setTime(new Date());
             pumpTime.setYear(calendar.get(Calendar.YEAR));
@@ -1141,10 +1142,13 @@ public class LocalInsightPlugin extends PluginBase implements PumpInterface, Con
 
     private void readHistory() {
         try {
-            PumpTime pumpTime = connectionService.requestMessage(new GetDateTimeMessage()).await().getPumpTime();
             String pumpSerial = connectionService.getPumpSystemIdentification().getSerialNumber();
+
+            // compute the delta between device and pump time so we can correct historyEvents to match device time
+            PumpTime pumpTime = connectionService.requestMessage(new GetDateTimeMessage()).await().getPumpTime();
             timeOffset = Calendar.getInstance(TimeZone.getTimeZone("UTC")).getTimeInMillis() - parseDate(pumpTime.getYear(),
                     pumpTime.getMonth(), pumpTime.getDay(), pumpTime.getHour(), pumpTime.getMinute(), pumpTime.getSecond());
+
             InsightHistoryOffset historyOffset = MainApp.getDbHelper().getInsightHistoryOffset(pumpSerial);
             try {
                 List<HistoryEvent> historyEvents = new ArrayList<>();
@@ -1165,7 +1169,7 @@ public class LocalInsightPlugin extends PluginBase implements PumpInterface, Con
                         historyEvents.addAll(newEvents);
                     }
                 }
-                Collections.sort(historyEvents);
+                Collections.sort(historyEvents); // sorts by eventPosition ascending, oldest event first
                 Collections.reverse(historyEvents);
                 if (historyOffset != null) processHistoryEvents(pumpSerial, historyEvents);
                 if (historyEvents.size() > 0) {
@@ -1197,12 +1201,25 @@ public class LocalInsightPlugin extends PluginBase implements PumpInterface, Con
     }
 
     private void processHistoryEvents(String serial, List<HistoryEvent> historyEvents) {
+        // this method expects historyEvents to be sorted from new to old in order to account for time changes.
+        // timeOffset is computed to be the current offset between device and pump and we will update this offset
+        // when we detect that the pump time has changed
         List<TemporaryBasal> temporaryBasals = new ArrayList<>();
         List<InsightPumpID> pumpStartedEvents = new ArrayList<>();
         for (HistoryEvent historyEvent : historyEvents)
             if (!processHistoryEvent(serial, temporaryBasals, pumpStartedEvents, historyEvent))
                 break;
         Collections.reverse(temporaryBasals);
+        for (int i = 0; i + 1 < temporaryBasals.size(); i++) {
+            TemporaryBasal current = temporaryBasals.get(i);
+            TemporaryBasal next = temporaryBasals.get(i + 1);
+            // remove cancel TBR records if they're immediately followed by another TBR record since
+            // they have no benefit and only one TBR in the same second can be saved to the db
+            if (current.durationInMinutes == 0 && Math.abs(next.date - current.date) < 1000) {
+                temporaryBasals.remove(i);
+                i--;
+            }
+        }
         for (InsightPumpID pumpID : pumpStartedEvents) {
             InsightPumpID stoppedEvent = MainApp.getDbHelper().getPumpStoppedEvent(pumpID.pumpSerial, pumpID.timestamp);
             if (stoppedEvent == null || stoppedEvent.eventType.equals("PumpPaused")) continue;
@@ -1537,6 +1554,7 @@ public class LocalInsightPlugin extends PluginBase implements PumpInterface, Con
         calendar.set(Calendar.HOUR_OF_DAY, hour);
         calendar.set(Calendar.MINUTE, minute);
         calendar.set(Calendar.SECOND, second);
+        calendar.set(Calendar.MILLISECOND, 0); // truncate leftover milliseconds from current UTC time
         return calendar.getTimeInMillis();
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1202,7 +1204,6 @@ private void processHistoryEvents(String serial, List<HistoryEvent> historyEvent
         for (HistoryEvent historyEvent : historyEvents)
             if (!processHistoryEvent(serial, temporaryBasals, pumpStartedEvents, historyEvent))
                 break;
-        Collections.reverse(temporaryBasals);
```
Suchiman: This `reverse` undid the `reverse` on line 1169 so we can just remove both.

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1216,7 +1217,6 @@ private void processHistoryEvents(String serial, List<HistoryEvent> historyEvent
             temporaryBasal.isAbsolute = false;
             temporaryBasals.add(temporaryBasal);
         }
-        Collections.sort(temporaryBasals, (o1, o2) -> (int) (o1.date - o2.date));
```
Suchiman: This was to bring the events back into order after appending the `TemporaryBasal` events synthesized from `PumpStarted` / `PumpStopped` events but past this point we're only saving `temporaryBasals` to the database (right below) which does not appear to be dependent on the order of the events (except for when they happen to occur in the same second which would cause an update)

Suchiman:
What i cannot explain is why this got so much worse with 2.6, i've blamed all the relevant methods and their changes were basically all <= january 2019.
https://github.com/MilosKozak/AndroidAPS/issues/2435#issuecomment-583888356 claims it happened in 2.5.1 too

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
TebbeUbben: Dangerous! The history events are processed in reverse order on purpose in order to take time changes into account. Your changes here can cause timestamps that are really off.

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
TebbeUbben: You're right, this doesn't do much, but I'd leave it for now because it makes investigating the database easier.

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1537,6 +1537,7 @@ private long parseDate(int year, int month, int day, int hour, int minute, int s
         calendar.set(Calendar.HOUR_OF_DAY, hour);
         calendar.set(Calendar.MINUTE, minute);
         calendar.set(Calendar.SECOND, second);
+        calendar.set(Calendar.MILLISECOND, 0); // truncate leftover milliseconds from current UTC time
```
TebbeUbben: Good spot, but I don't think it has caused any trouble.

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
TebbeUbben: Actually, the only ways I see to solve this problem are:

1. Don't insert zero-duration TBRs into the database at all, but adjust the duration field of the existing one. AndroidAPS doesn't allow this at the moment, however, changing that behavior make cause problems with Nightscout again.
2. Add some artificial delay to start or stop entries to make sure that they appear in the right order, however, this could mess everything up in some other way.

TebbeUbben:
Actually, the only ways I see to solve this problem are:

1. Don't insert zero-duration TBRs into the database at all, but adjust the duration field of the existing one. AndroidAPS doesn't allow this at the moment, however, changing that behavior make cause problems with Nightscout again.
2. Add some artificial delay to start or stop entries to make sure that they appear in the right order, however, this could mess everything up in some other way.

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
Suchiman: I think this + the processing in revers order was responsible for this working half of the time (when either one record managed to round up to the next second or "overflowing" + the sort by date just before inserting fixed the override order, but i didn't trace it that far)

Suchiman:
Hm... back to the drawing board i guess...

1: I wonder if adjusting duration in the db is necessary
Screenshot_20200323-035936
But yeah... i only tested with ns upload only, who knows what will happen with bidirectional sync.

Yeah i've considered 2, i would have added a second pass over the final set of `temporaryBasals` and subtracted a second from the "cancel TBR" events... but in either direction (+ -) you could stumble into an (db) existing event...

TebbeUbben:
> 1: I wonder if adjusting duration in the db is necessary

What you see is because of how Android processes TBRs. When it sees a one that starts in the middle of another, the older one will be cut off. See [NonOverlappingIntervals](https://github.com/MilosKozak/AndroidAPS/blob/1ab004a39ec26eab2ef55ec31039a46037a51894/app/src/main/java/info/nightscout/androidaps/data/NonOverlappingIntervals.java). Adjusting existing TBRs instead of creating a stop TBR removes that odd edge causing all the trouble.

chrisbev:
May be silly but just a thought. If problem starts with Insight resolution of 1second and not AAPS resolution is it possible in AAPS to add crude extra resolution to the Insight times to get the right order? Also the random miliseconds mensioned above are they a fixed number for a while or different each time. If fixed for each pump/AAPS that might bias some pumps to be more affected than others? If I'm way off don't waste time with too much typing sometimes a random comment spurs a real solution?

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
Suchiman: To be more precise, when processing events backwards, this lead to the newer (start TBR) events usually having smaller milliseconds so same second events would flip around once they got sorted by date just before insert, which would greatly increase the chance that a cancel TBR record would override the start TBR record.

Suchiman:
@chrisbev AAPS uses unix time milliseconds but for the internal database, it will (sometimes) round to seconds and then check if a TBR that started on the same seconds already exists, so AAPS can only save one TBR per second.

@TebbeUbben I've went for the most conservative fix, i'll remove a cancel TBR record when there's a start TBR record following in less than a second, so it will not cause conflicting updates. Nightscout will also not receive the cancel TBR record and won't do anything stupid. And as far as AAPS is concerned, starting another TBR is as good as an cancellation record so adjusting duration is only a nice to have. If a TBR actually gets canceled without starting a new TBR, we still record a cancel event. [AutoTune for example](https://github.com/openaps/oref0/blob/acb0150021e7bf284b6741a9510076a13f690747/lib/iob/history.js#L383) computes the actual duration and then discards the cancel events anyway.

I've just now noticed though, that while we will not save the `TemporaryBasal` record, we do however save the `InsightPumpID` of that event, but that shouldn't be an issue?

DigitalDan1:
Hi @Suchiman,

I resolved this issue in a different way, basically by including the TBR duration in the sorting, and not overwriting existing records with pumpId: https://github.com/MilosKozak/AndroidAPS/pull/2249 

This has been working well for about 4 months.  I'd be interested in your opinion.

app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
DigitalDan1: I think there are a couple more places this fix should be applied:

https://github.com/MilosKozak/AndroidAPS/blob/0f6f344f6307cbd5765e01a2355314aa204e6ec2/app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java#L356-L363

and

https://github.com/MilosKozak/AndroidAPS/blob/0f6f344f6307cbd5765e01a2355314aa204e6ec2/app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java#L1565-L1574

Suchiman:
@TebbeUbben please advise on target branch.
I've been running with these changes for over a month now on my rig and data shown in AAPS and nightscout matches up and looks correct.

PR #2534, opened 2020-03-25
Support for master password

dlvoy:
Support for master password 
- storing password as hashes (HMAC) instead of plaintext
- additional crypto utils with tests _(partialy for in-progress pref export encryption)_,
- changed PasswordCheck to more common UI and to use lambdas _(will need given password in lambda callback for prefs enc soon)_

not done:
- complexity check on passwords
- adding master password setup into configuration wizard (do we need it at this point)

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/activities/MyPreferenceFragment.kt
+++ b/app/src/main/java/info/nightscout/androidaps/activities/MyPreferenceFragment.kt
@@ -50,8 +50,10 @@ import info.nightscout.androidaps.plugins.source.EversensePlugin
 import info.nightscout.androidaps.plugins.source.GlimpPlugin
 import info.nightscout.androidaps.plugins.source.PoctechPlugin
 import info.nightscout.androidaps.plugins.source.TomatoPlugin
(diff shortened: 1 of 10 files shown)


guydavies:
Hi Daniel

This looks really interesting. One note. Please DON'T create complexity
rules for passwords. They're utterly discredited as a mechanism for
password strength because they tend to make people write down the password
(ref NIST).

Better would be to ensure that password managers can correctly interact
with the system or even setup 2FA and/or OTP.

Rgds

Guy

dlvoy:
Thanks for that! As password manager user - agree!
I will add `password.setAutofillHints(View.AUTOFILL_HINT_PASSWORD);` 
it is API dependent so will be only conditionally available

_And I am Dominik ;) but it is common mistake :D_

guydavies:
Yes. I noticed my mistake but didn't want to just generate noise with an
apology 😁.

Thanks for taking the suggestion and the great response.

Rgds

Guy

dlvoy:
Screenshot_1585144171
Screenshot_1585144200
Screenshot_1585144221

PR #2550, opened 2020-04-05
Autosens Upgrade

Tornado-Tim:
This Pull Request brings Autosens into line with the Openaps implementation.

That being Autosens is ran under a 24 hour window and 8 hour window, whichever of these is the most sensitive it will use.

What this PR does:
It removes oref0 sensitivity in its entirety, as Autosens does not use 24 alone and reimplements it in oref1.

Exam answers have been adjusted and redundant text assets removed.

What is still needed?
A translation of "description_sensitivity_oref1" variable for every language.


MilosKozak:
don't touch other than english strings otherwise it leads to conflicts.
Non EN string will be removed by Crowdin

PR #2553, opened 2020-04-05
Introduced Patient name, refactored SMS OTP and Prefs Export to use it

dlvoy:
* Added _Patient name_ as general preference
* Refactored SMS OTP to use it for generated QRCode as name of authenticator subject / "account" in AAPS (instead of separate filed in plugin configuration)
* Used in preferences export
* Added to configuration wizard, before _Patient age_


app/src/main/java/info/nightscout/androidaps/plugins/general/smsCommunicator/otp/OneTimePassword.kt
```diff
@@ -48,8 +48,8 @@ class OneTimePassword @Inject constructor(
      * Name of master device (target of OTP)
      */
     fun name(): String {
-        val defaultUserName = resourceHelper.gs(R.string.smscommunicator_default_user_display_name)
-        var userName = sp.getString(R.string.key_smscommunicator_otp_name, defaultUserName).replace(":", "").trim()
+        val defaultUserName = resourceHelper.gs(R.string.patient_name_default)
+        var userName = sp.getString(R.string.key_patient_name, defaultUserName).replace(":", "").trim()
         if (userName.isEmpty())
             userName = defaultUserName
         return userName
```
MilosKozak, line 55:
```
fun name(): String  =
   sp.getString(R.string.key_patient_name, defaultUserName).replace(":", "").trim().let {
  if (it.isEmpty()) resourceHelper.gs(R.string.patient_name_default) else it
}
```
;)

PR #2564, opened 2020-04-11
Force Short Text in Status light (proposal)

Philoul:
It's just a proposal: Long text for status lights are too long in a phone in portrait mode (even in english it's too long, in french it's worse)
Here I forced short text, but may be if you want long text for tablets you could change how you switch between short and long text...

StatusLight_Short


MilosKozak:
status lights are refactored in 2.6.5-dev
instead of 3 different pieces of code now is only one general
please test and create PR to that if needed

PR #2565, opened 2020-04-11
Fix requesting Dex permission with Android 10.

jotomo:


jotomo:
... nope, that's not what's causing it.

PR #2602, opened 2020-04-27
WIP: Add skip Neutral Temps preference

Tornado-Tim:
Adds a preference so Neutral Temps can be turned off or on, rather than being set by default on.

Can adjust to make this a preference for some pumps not others, just need feedback on this direction.


app/src/main/res/values/strings.xml
```diff
@@ -956,6 +956,9 @@
     <string name="key_enableSMB_with_temptarget" translatable="false">enableSMB_with_temptarget</string>
     <string name="key_enableSMB_after_carbs" translatable="false">enableSMB_after_carbs</string>
     <string name="key_allowSMB_with_high_temptarget" translatable="false">enableSMB_with_high_temptarget</string>
+    <string name="key_skip_neutral_temps" translatable="false">skip_neutral_temps</string>
+    <string name="skip_neutral_temps_title">Skip neutral temp basals</string>
+    <string name="skip_neutral_temps_summary">If enabled, it skips setting basal to neutral before the end of each hour. This method can help stop some pumps beeping/vibrating on the hour.</string>
```
MilosKozak, line 961: which pumps?

MilosKozak:
i've no real opinion as i'm not MDT user

app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/DetermineBasalAdapterSMBJS.java
```diff
@@ -265,7 +269,11 @@ public void setData(Profile profile,
         mProfile.put("exercise_mode", SMBDefaults.exercise_mode);
         mProfile.put("half_basal_exercise_target", SMBDefaults.half_basal_exercise_target);
         mProfile.put("maxCOB", SMBDefaults.maxCOB);
-        mProfile.put("skip_neutral_temps", SMBDefaults.skip_neutral_temps);
+        if (!manufacturer.name().equals("Medtronic")) {
```
AdrianLxM, line 272: Imho this adapter should be pump-manufacturer agnostic.
I'd add a new function `shouldStopTbrAtFullHourIfSave()` (might be better names) into pump interface with a default to false. Then this logic could be part of the Medtronic plugin.

app/src/main/assets/OpenAPSSMB/determine-basal.js
```diff
@@ -897,7 +897,8 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
 
     // if not in LGS mode, cancel temps before the top of the hour to reduce beeping/vibration
     // console.error(profile.skip_neutral_temps, rT.deliverAt.getMinutes());
-    if ( profile.skip_neutral_temps && rT.deliverAt.getMinutes() >= 55 ) {
+    //patch determine basal so skip_neutral_temps works as expected
+    if ( !profile.skip_neutral_temps && rT.deliverAt.getMinutes() >= 55 ) {
```
AdrianLxM, line 901: The actual logic of `skip_neutral_temps` is in `basal-set-temp.js`.

A "neutral temp" is a TBR that is *not* cancelled despite being the same as the currently programmed basal rate.

"skip_neutral_temps -> cancel TBR if close to the one in the profile" could be seen here as well, so this change would negate the logic.

I'd suggest either have a second flag - or for the sake of having the same `determine-basal.js` as OpenAPS:
Have it set to `false` per default for all pumps and only the Medtronic expose it as true?

Other pumps cannot set neutral temps -> they will fallback to skipping them anyway.

app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/DetermineBasalAdapterSMBJS.java
Tornado-Tim: I got asked by Milos to do this. So ???????

MilosKozak:
i thought @AdrianLxM told i did it oposite way

Tornado-Tim:
@MilosKozak this seems to work as I thought it would now. @AdrianLxM can you confirm?

AdrianLxM:
The logic seems to be inverted.

"If enabled, it sets basal to neutral before the end of each hour. This method can help stop some pumps beeping/vibrating on the hour."

A neutral temp will vibrate though. Like any temp.
A cancelled temp will not.

AdrianLxM:
I would therefore to avoid the term "neutral temp" in the strings of the setting to avoid confusion.

Tornado-Tim:
@AdrianLxM updated the text, sufficient?

PR #2605, opened 2020-04-28
RandomBGPlugin - stay between min and max

AdrianLxM:
The sinus is in interval [-1,1].
If we shift it by one interval it would map it to 0 to 2 times the interval -> divide by two.


MilosKozak:
merged through dev

PR #2610, opened 2020-05-01
Add an explicit Low Glucose Suspend mode

Tornado-Tim:
This PR makes it possible for people to enter into Low Glucose Suspend when they wish without needing to undo Objective 6 to do it.

When Low Glucose Suspend (LGS) is enabled the Main APS menu will state it is in such a state.
When Objective 6 is enforced or the Max IOB is set to 0, the Main UI will reflect as such.

To change directly into LGS the user would change modes via the APS menu.
Main UI will look as follows when LGS is in effect.
Screenshot (1 May 2020 22_12_10)


app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
```diff
@@ -622,8 +622,10 @@ class OverviewFragment : DaggerFragment(), View.OnClickListener, OnLongClickList
                 }
 
                 loopPlugin.isEnabled(PluginType.LOOP)                            -> {
-                    overview_apsmode?.text = if (closedLoopEnabled.value()) resourceHelper.gs(R.string.closedloop) else resourceHelper.gs(R.string.openloop)
-                    overview_apsmode?.setBackgroundColor(resourceHelper.gc(R.color.ribbonDefault))
+                    val APSmode = sp.getString(R.string.key_aps_mode, "open")
+                    val isLGS = loopPlugin.isLGS
+                    overview_apsmode?.text = if (closedLoopEnabled.value()) if (isLGS) resourceHelper.gs(R.string.lgs) else  resourceHelper.gs(R.string.closedloop) else resourceHelper.gs(R.string.openloop)
```
AdrianLxM, line 627: I find multiple if/else in one line and without `{}` hard to read and it is easy to introduce bugs in the future.
In Kotlin even multi-line if/else cascades are expressions -> I'd rather have a few more lines here.

app/src/main/java/info/nightscout/androidaps/utils/HardLimits.kt
```diff
@@ -68,6 +72,10 @@ class HardLimits @Inject constructor(
         return MAXIOB_SMB[loadAge()]
     }
 
+    fun maxIobLGS(): Double {
+        return MAXIOB_LGS
+    }
+
```
AdrianLxM, line 78: As it is a `val` property, it is read-only (with auto-generated getter-function when calling from Java code).  We don't need a separate getter function.

Some other's use a getter function as they are dependent on age - so dynamic.

app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
```diff
@@ -288,6 +288,21 @@ public boolean isSuspended() {
         return true;
     }
 
+    public boolean isLGS(){
+        Constraint<Boolean> closedLoopEnabled = constraintChecker.isClosedLoopAllowed();
+        Double MaxIOBallowed = constraintChecker.getMaxIOBAllowed().value();
+        String APSmode = sp.getString(R.string.key_aps_mode, "open");
+        Double LGSthreshold = 0d;
```
AdrianLxM, line 295: Shouldn't this come from the `HardLimits.kt` constants?

Tornado-Tim:
Updated PR with fixes, thanks Adrian, much appreciated :)

MilosKozak:
tests must be updated

Tornado-Tim:
> tests must be updated

Fixed tests

MilosKozak:
still failing

PR #2627, opened 2020-05-07
Temp Basal new icons for temp basal in overview

Philoul:


swissalpine:
Are you sure? The action tab shows a possible action (x = cancel a running tbr), the overview tab the actual status (missing x = running tbr)

guydavies:
yes. but in the new overview page, that's reversed. if the temp is running, it shows no X. if a temp isn't running, the X is shown. to me, that's the wrong way round. if that's what @Philoul  is fixing,. I agree with the fix.

swissalpine:
Yes, I know. And I think it's correct as it is. This are different pages and (unfortunatly) the same symbol has different meanings (as written above): On action tab an action (x signals that I can stop the tbr), on overview tab a status (x symbols there is no tbr running).

guydavies:
nah. that's just broken. it's the same thing (is there a TBR active or not). it should be consistent.

swissalpine:
Sorry, but you can't convince me. We will see if this will be "fixed" ...

Philoul:
My suggestion is to not use exactly the same icons to avoid misunderstanding:
See below 2 proposals made quickly this morning without the "x" (on the left with 2 icons, on the right with 3 icons)...
TBR Icon

For @MilosKozak if ok with one of these 2 proposal I can make clean icons and integrate this in code...

Philoul:
A third proposal on the left... (Straight line if no TBR, no X to avoid inconsitencies and misunderstanding with action tab, and we can easily see if it's a TBR below or above 100%...).

I like the third one because icons are very close to actions icon so I think it's easier to understand...

TBR Icon

guydavies:
I like the third one. It clearly shows that there's a TBR compared to the
regular "flat" basal. +1 for that.

Philoul:
New icons created with third proposal
=> I also added my photoshop files in /icons folder

Tested on my A40 (high res 1080x2340px) and on my dev phone (low res 480x960px)

See screenshot of my dev phone with high TBR and loop phone with zero-temp

Screenshot_20200508-181940

Screenshot_20200508-182918_AndroidAPS

Issue #2626, opened 2020-05-06
Enable/disable insight vibration on bolus


JimZam:
Good evening. I see Disable Insight vibration on bolus as an advantage but I would like to know if it's possible to enable/disable it. Sometimes the vibrations alert me AAPS is doing something and I check why just to be sure everything is fine.
Thank you so much.
Kind regards

MilosKozak:
it's hardcoded

PR #2647, opened 2020-05-15
Add vibration in Insight preferences

Philoul:
I think it could be a good idea to add this feature in Insight settings... (see https://github.com/MilosKozak/AndroidAPS/issues/2626)
Build works but as it related to pump communication, I prefer to not test it on my pump before Tebbe confirmation... (code is very simple, so it should work!)
(This setting is only in pref_insight_local_full, I didn't update pref_insight_local_pumpcontrol...)

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
@@ -597,7 +597,7 @@ public class LocalInsightPlugin extends PumpPluginBase implements PumpInterface,
                     bolusMessage.setDuration(0);
                     bolusMessage.setExtendedAmount(0);
                     bolusMessage.setImmediateAmount(insulin);
-                    bolusMessage.setVibration(sp.getBoolean("insight_disable_vibration",false));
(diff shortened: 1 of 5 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -820,6 +821,7 @@ public PumpEnactResult setExtendedBolusOnly(Double insulin, Integer durationInMi
             bolusMessage.setDuration(durationInMinutes);
             bolusMessage.setExtendedAmount(insulin);
             bolusMessage.setImmediateAmount(0);
+            bolusMessage.setVibration(sp.getBoolean("insight_disable_vibration",false));
```
TebbeUbben, line 824: You could also add a `boolean` parameter `disableVibration` to the method that you set to false if called from `setExtendedBolus` and true if called from `setTempBasalAbsolute` (TBR emulation) and vibration is enabled for automated treatments.

app/src/main/res/xml/pref_insight_local_full.xml
```diff
@@ -48,6 +48,12 @@
             android:summary="@string/enable_tbr_emulation_summary"
             android:title="@string/enable_tbr_emulation" />
 
+        <SwitchPreference
```
TebbeUbben, line 51: Please add this to pumpcontrol build flavour, too.

TebbeUbben:
@MilosKozak Please merge.

PR #2655, opened 2020-05-17
Fix Wear Control

Philoul:
Kotlin doesn't like regex... (AAPS crashed on these lines with ""\\s+", replaced by a space)
It may be less reliable than regex but in my memories there is always a single space between words in commands sent by Wear app...
Tested with a bolus and pump status and ok


Philoul:
I think my second commit is better than first one because we still have regex for command split...

PR #2688, opened 2020-05-28
Add translation for secondary graph title

Philoul:
Everything is in title...


MilosKozak:
i didn't add this to translations because it could be challenging in many languages
@AdrianLxM @TebbeUbben ?

Philoul:
I proposed this PR only because we use IA and GA for IOB and COB in french...
I didn't planned to translate the other curves (ACT, SEN, ABS, DEV) in french.

PR #2691, opened 2020-05-30
ABS Curve in background

Philoul:
I didn't change order in selection menu (otherwise selected curves are changed in secondary graphs...)


Philoul:
But answering last issue of Guy Davies, I think it's necessary to also change menu for consistency with the numerical scale on the left...

guydavies:
hi @philoul look at the scale on both. it's the same. but the final hour shows a negligible change to sensitivity in the broader view but a change of ~6 in the narrower view because the bigger change earlier in the day is now obscured. it's not the scales that are an issue. it's the actual displayed sensitivity variation.

guydavies:
to be honest. the numbers at the left are utterly meaningless because they have no units and refer to all the different traces. they're actually confusing rather than informative.

MilosKozak:
numbers on left are valid for data where is the scale calculated from
a bit mess if you turn on many many elements in one graph, right, but useful if you you have one element selected

PR #2693, opened 2020-05-31
BugFix set min SMB frequency number to 3

Tornado-Tim:
Since AAPS has a defined limit that will not deliver SMB's less than 3 min apart, there is not point in being able to set a number lower than that as it will not work.  This fixes this.

We can lower this value later when these restrictions can be removed and when more frequently updated CGM delta points become more the normal.


MilosKozak:
👍

PR #2694, opened 2020-05-31
Visually Disable SMB preferences based on objectives

Tornado-Tim:
Visually it can be a bit confusing to people to have the ability to turn preferences on/off even though it doesn't work if the objectives have not passed.

This PR adds the ability to disable/re-enable autosens and SMB preferences based on objectives.

This makes it more clear for the user which things can be set or not.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/activities/MyPreferenceFragment.kt
+++ b/app/src/main/java/info/nightscout/androidaps/activities/MyPreferenceFragment.kt
@@ -333,6 +333,35 @@ class MyPreferenceFragment : PreferenceFragmentCompat(), OnSharedPreferenceChang
                     if (pref != null) pref.isEnabled = true
         }
 
+        if (p.getKey() == resourceHelper.gs(R.string.key_smbinterval) ) {
+            val pref: Preference? = findPreference(resourceHelper.gs(R.string.key_smbinterval))
+            val c =  constraintChecker.isSMBModeEnabled()
+            if (!c.value())
+                if (pref != null) pref.isEnabled = false
+                else
+                    if (pref != null) pref.isEnabled = true
+        }
+
+        if (p.getKey() == resourceHelper.gs(R.string.key_smbmaxminutes) ) {
+            val pref: Preference? = findPreference(resourceHelper.gs(R.string.key_smbmaxminutes))
+            val c =  constraintChecker.isSMBModeEnabled()
+            if (!c.value())
+                if (pref != null) pref.isEnabled = false
+                else
+                    if (pref != null) pref.isEnabled = true
+        }
+
+        if (p.getKey() == resourceHelper.gs(R.string.key_uamsmbmaxminutes) ) {
+            val pref: Preference? = findPreference(resourceHelper.gs(R.string.key_uamsmbmaxminutes))
+            val c =  constraintChecker.isSMBModeEnabled()
+            if (!c.value())
+                if (pref != null) pref.isEnabled = false
+                else
+                    if (pref != null) pref.isEnabled = true
+        }
+
+
+
     }
 
     // We use Preference and custom editor instead of EditTextPreference
--- a/app/src/main/java/info/nightscout/androidaps/activities/MyPreferenceFragment.kt
+++ b/app/src/main/java/info/nightscout/androidaps/activities/MyPreferenceFragment.kt
@@ -24,6 +24,7 @@ import info.nightscout.androidaps.plugins.bus.RxBusWrapper
 import info.nightscout.androidaps.plugins.configBuilder.PluginStore
 import info.nightscout.androidaps.interfaces.ProfileFunction
 import info.nightscout.androidaps.plugins.constraints.safety.SafetyPlugin
+import info.nightscout.androidaps.plugins.configBuilder.ConstraintChecker
 import info.nightscout.androidaps.plugins.general.automation.AutomationPlugin
 import info.nightscout.androidaps.plugins.general.maintenance.MaintenancePlugin
 import info.nightscout.androidaps.plugins.general.nsclient.NSClientPlugin
@@ -66,6 +67,8 @@ class MyPreferenceFragment : PreferenceFragmentCompat(), OnSharedPreferenceChang
     @Inject lateinit var profileFunction: ProfileFunction
     @Inject lateinit var pluginStore: PluginStore
     @Inject lateinit var config: Config
+
+    @Inject lateinit var constraintChecker: ConstraintChecker
 
     @Inject lateinit var automationPlugin: AutomationPlugin
     @Inject lateinit var danaRPlugin: DanaRPlugin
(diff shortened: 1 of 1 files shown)


MilosKozak:
i'd move this preprocessing to plugin where it belongs to (`PluginBase::preprocessPreferences`) ... easier to maintain

app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/OpenAPSSMBPlugin.java
```diff
@@ -129,6 +132,39 @@ public long getLastAPSRun() {
...
+            if (!cSMB.value())
+                if (pref != null) pref.setEnabled(false);
+                else
+                if (pref != null) pref.setEnabled(true);
+        }
+
+        Preference pref = preferenceFragment.findPreference(resourceHelper.gs(R.string.key_openapsama_useautosens));
+        if (!cAMA.value())
+            if (pref != null) pref.setEnabled(false);
+            else
```
MilosKozak, line 163: wrong formating or meaning?

app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/OpenAPSSMBPlugin.java
```diff
@@ -129,6 +132,39 @@ public long getLastAPSRun() {
...
+        };
+
+        Constraint<Boolean> cSMB = constraintChecker.isSMBModeEnabled();
+        Constraint<Boolean> cAMA = constraintChecker.isAMAModeEnabled();
+
+        for (int i=0; i<=7; i++) {
+            Preference pref = preferenceFragment.findPreference(resourceHelper.gs(smbprefkeys[i]));
+            if (!cSMB.value())
+                if (pref != null) pref.setEnabled(false);
+                else
```
MilosKozak, line 156: wrong formating or meaning?

app/src/main/java/info/nightscout/androidaps/plugins/constraints/safety/SafetyPlugin.java
```diff
@@ -128,9 +128,6 @@ public SafetyPlugin(
 
     @NonNull @Override
     public Constraint<Boolean> isSMBModeEnabled(@NonNull Constraint<Boolean> value) {
-        boolean enabled = sp.getBoolean(R.string.key_use_smb, false);
-        if (!enabled)
-            value.set(getAapsLogger(), false, getResourceHelper().gs(R.string.smbdisabledinpreferences), this);
```
MilosKozak, line 133: but this removes message passed to UI

Tornado-Tim: Yes, unfortunately its required because the constraint will be raised for `isSMBModeEnabled()` stating that a constraint was raised just because the preference was off.
And this means the preferences will be locked and cant be turned on because they are locked because the constraint says SMB is off.

The constraint doesn't enforce safety it just states SMB is off.
If we can report that its off without setting the constraint then this will work.

Tornado-Tim:
Made the code logic a bit more clear with the brackets in place.

app/src/main/java/info/nightscout/androidaps/plugins/constraints/safety/SafetyPlugin.java
Tornado-Tim: OK fixed in the latest commits

MilosKozak: result is passed to determine_basal and used there.
I'm really not sure we can simply remove it

app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/OpenAPSSMBPlugin.java
```diff
@@ -129,6 +132,40 @@ public long getLastAPSRun() {
...
+                R.string.key_enableSMB_with_COB,
+                R.string.key_enableSMB_with_temptarget,
+                R.string.key_allowSMB_with_high_temptarget,
+                R.string.key_enableSMB_after_carbs
+        };
+
+        Constraint<Boolean> cSMB = constraintChecker.isSMBModeEnabled();
+        Constraint<Boolean> cAMA = constraintChecker.isAMAModeEnabled();
+
+        for (int i=0; i<=7; i++) {
```
AdrianLxM, line 152: Here I'd use `< smbprefkeys.length` (or `size()` depending on the datastrucutre), not a hardcoded 7.

AdrianLxM:
I know that Milos wanted it to be in the plugin...
... but doen't this mean it will only be effective if you open the preference for the plugin?

Or is it also active if you use the general preferences that accumulates all preferences?

app/src/test/java/info/nightscout/androidaps/interfaces/ConstraintsCheckerTest.kt
```diff
@@ -210,7 +210,7 @@ class ConstraintsCheckerTest : TestBaseWithProfile() {
 //        `when`(constraintChecker.isClosedLoopAllowed()).thenReturn(Constraint(true))
         val c = constraintChecker.isSMBModeEnabled()
         Assert.assertEquals(true, c.reasonList.size == 3) // 2x Safety & Objectives
-        Assert.assertEquals(true, c.mostLimitedReasonList.size == 3) // 2x Safety & Objectives
+        Assert.assertEquals(true, c.mostLimitedReasonList.size == 2) // 2x Safety & Objectives
```
AdrianLxM, line 213: Here the test now asserts 2, the comment suggests 3.

app/src/test/java/info/nightscout/androidaps/plugins/constraints/safety/SafetyPluginTest.kt
```diff
@@ -110,7 +110,7 @@ class SafetyPluginTest : TestBaseWithProfile() {
         var c = Constraint(true)
         c = safetyPlugin.isSMBModeEnabled(c)
         Assert.assertTrue(c.getReasons(aapsLogger).contains("SMB disabled in preferences"))
-        Assert.assertEquals(false, c.value())
+        c.value()
```
AdrianLxM, line 113: Here the actual Assertion is missing.

MilosKozak:
preprocessing is done always not only while plugin is active

app/src/main/java/info/nightscout/androidaps/plugins/constraints/safety/SafetyPlugin.java
AdrianLxM: Determine basal has a parameter to allow SMB. In `info/nightscout/androidaps/plugins/aps/openAPSSMB/DetermineBasalAdapterSMBJS.java`it is in line 150: `new Boolean(mMicrobolusAllowed)`

This is set via `setData()` called from `info/nightscout/androidaps/plugins/aps/openAPSSMB/OpenAPSSMBPlugin.java` which uses the ConstraintChecker to determine the value.

-> The ConstraintChecker either needs to handle this... or it needs to be handled separately before calling determine_basal.

app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/OpenAPSSMBPlugin.java
Tornado-Tim: I have just become accustomed to hard coding them if the array size is known and not going to change as there is a slight performance penalty with asking how big a data structure is.

Tornado-Tim:
Hi @MilosKozak and @AdrianLxM in the latest commits I have changed the method of approach here as it seems the isSMBEnabled constraint was not the correct one to use as its tied to turning on and off via preference.
So an new constraint has been made isSMBAllowed.

app/src/test/java/info/nightscout/androidaps/interfaces/ConstraintsCheckerTest.kt
AdrianLxM: Here still is a mismatch between comment and assertion.

zainebukra:
@AdrianLxM Hi, are you able to please help me with getting nightwatch to work for me (from New Zealand)? :)

PR #2709, opened 2020-06-08
Add toTimeMinutesFromMidnight to DateUtil

Philoul:
I propose to add this function to DateUtil (I don't find equivalent in existing function...)
```public static long toTimeMinutesFromMidnight(long timeStamp, int minutesFromMidnight)```
It returns a time offset from midnight the day of timeStamp
i.e. 
```
DateUtil.toTimeMinutesFromMidnight(System.currentTimeMillis(), 120)
```
return a long value for today at 2AM (local time)

I use this function in AutotunePlugin...


MilosKozak:
MidnightTime.class

Philoul:
Ok, thank's

PR #2710, opened 2020-06-09
Constraint IsClosedLoopAllowed -> IsClosedLoopEnabled

Tornado-Tim:
PR to allow the following constraints `IsClosedLoopAllowed` to `IsClosedLoopEnabled` follow what they say.

Implementation:  `IsClosedLoopAllowed` Every existing check that used `IsClosedLoopEnabled` previously minus preference checking.
`IsClosedLoopEnabled` same constraints as `IsClosedLoopAllowed` plus checking/using if the closed loop preference as a constraint.

This allows the `IsClosedLoopAllowed` constraint to see if a user is eligible for closed loop.


MilosKozak:
there is so much code on so many places (mostly duplicated) that finally I'm missing the point why the code exists

PR #2726, opened 2020-06-17
Add loop shortcuts to overview.

Tornado-Tim:
This allows to switch between APS closed and open modes on the main screen (if passed objectives and constraints).


MilosKozak:
i don't like the dependency isClosedLoopEnabled -> isClosedLoopAllowed. I don't have any specific reason now. it's jut a bad feeling about possible issues in the future and necessity add this dependency to every constraint plugin overriding isClosedLoopEnabled.
But I like the idea of switching open-closed loop mode from overview.
So. I'd keep only isClosedLoopEnabled and customize overview menu based on shared preference value only.

Tornado-Tim:
> i don't like the dependency isClosedLoopEnabled -> isClosedLoopAllowed. I don't have any specific reason now. it's jut a bad feeling about possible issues in the future and necessity add this dependency to every constraint plugin overriding isClosedLoopEnabled.
> But I like the idea of switching open-closed loop mode from overview.
> So. I'd keep only isClosedLoopEnabled and customize overview menu based on shared preference value only.

Hey @MilosKozak , unfortunately this is not possible as its necessary to determine if closed loop is allowed vs Enabled. I cannot check isClosedLoopEnabled constraint as this value will be false as soon as Openloop is turned on or forced on by the Objective constraints.

This is why I had to separate the two, Enabled vs Allowed.

MilosKozak:
my proposal
https://github.com/MilosKozak/AndroidAPS/compare/menuswitch

Tornado-Tim:
This 

> > i don't like the dependency isClosedLoopEnabled -> isClosedLoopAllowed. I don't have any specific reason now. it's jut a bad feeling about possible issues in the future and necessity add this dependency to every constraint plugin overriding isClosedLoopEnabled.
> > But I like the idea of switching open-closed loop mode from overview.
> > So. I'd keep only isClosedLoopEnabled and customize overview menu based on shared preference value only.
> 
> Hey @MilosKozak , unfortunately this is not possible as its necessary to determine if closed loop is allowed vs Enabled. I cannot check isClosedLoopEnabled constraint as this value will be false as soon as Openloop is turned on or forced on by the Objective constraints.
> 
> This is why I had to separate the two, Enabled vs Allowed.

This still allows closed loop even when an objective hasnt passed. The checkbox breaks UI convention as well.
Screenshot (18 Jun 2020 08_37_34)

MilosKozak:
I know. But i'd not break code logic in favor of one UI element.
Maybe renaming to "Prefer closed loop" ?
another opinion @andyrozman @AdrianLxM @TebbeUbben ?

Tornado-Tim:
> I know. But i'd not break code logic in favor of one UI element.
> Maybe renaming to "Prefer closed loop" ?
> another opinion @andyrozman @AdrianLxM @TebbeUbben ?

This is the thing, these additional constraints will be used in future to dynamically modify the UI based on what is possible to the user. In this PR it just one item, but can apply to more items. This is similar what I am trying to achieve with #2694, eliminate the confusion about what options are available with the objective they are up to.

Currently we can have options set to something but they don't do anything and this confuses people, by visually restricting what is possible the confusion to the user is eliminated, and helps show the user when the settings get opened up more as objectives are completed.

guydavies:
Hi Milos,

I’m not a programmer, so I am not clear on the implications of introducing this change, but I cannot fault Tim’s logic.  If the menus are to change dynamically according to what is already achieved in the Objectives, then IsthisPermitted is logically different from IsThisEnabled.  The former is simply an identifier of a specific function’s accessibility according to where the user is in the Objectives.  The latter is whether the user chooses to enable that feature/function.  Overloading one of those values with both meanings is risky.

As I say, this is from a purely logical perspective.  I admit that it may introduce greater complexity into the code.

Rgds,

Guy

PR #2732, opened 2020-06-18
Don't check the hypo TT checkbox on another TT

PoweRGbg:
Don't check hypo tt checkbox when another temp is already active

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/dialogs/CarbsDialog.kt
+++ b/app/src/main/java/info/nightscout/androidaps/dialogs/CarbsDialog.kt
@@ -122,8 +122,9 @@ class CarbsDialog : DialogFragmentWithDate() {
             validateInputs()
         }
 
+        val tempTarget = treatmentsPlugin.tempTargetFromHistory
         iobCobCalculatorPlugin.actualBg()?.let { bgReading ->
-            if (bgReading.value < 72)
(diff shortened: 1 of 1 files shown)


MilosKozak:
it's not a good idea when there is low TT running

PoweRGbg:
For me - when there is already TT running, there is something/someone who started it and there is a reason. I'm sick of my automation rule setting hypo TT, but when I try to add carbs it sets another one. 
I'm not removing the ability to set a new one, just the auto tick.

PoweRGbg:
@MilosKozak  Is that what you meant?

MilosKozak:
@AdrianLxM @TebbeUbben 
I'm still not sure it's a good idea. Previous TT may end in a few minutes

AdrianLxM:
We would have to check that the TT running is at least as high and at least as long than the hypo TT.

Then again setting the HypoTT on top of another won't hurt. -> I'd keep it as is.

guydavies:
Hmm. I'm not sure I agree that setting the Hypo TT can't hurt. If I already
have activity TT active which is much higher than the Hypo TT, replacing
the Activity TT with the Hypo TT would definitely increase the risk (not a
lot but it would be more of a risk) that the algorithms would not catch
things. I know I should stop exercise when I get that but if I'm doing
weights (raises BG) after doing cardio (drops BG) then continuing also
helps in my experience.

PR #2741, opened 2020-06-19
Vector icons

Philoul:
Now most of png icons are replaced
Still remain cleanings (several icons are identical or slightly scaled with different file names)
It will also be necessary to harmonize the colors with the color palette (4 or 5 different green or red, ...)

See below new vector icons
NewVector2


chrisbev:
Hi, well done, if any of the colours are paramount, rather than nice to have, have you considered someone who is colour blind?

guydavies:
That’s actually an important reason why having a single icon with different
colours to represent different states is a bad thing. Different situations
need different icons (those different icons may have different colours but
don’t assume that a variation of colours is sufficient). Also, the variety
of different colour-blindness makes it difficult to pick colour
combinations that work for all.

Rgds

Guy

Philoul:
I didn't change anything, just replaced bitmap files by vector, so color and icons are "as is".
according to your screen size you won't see any change or maybe a little sharper.
Next steps (for futur release) could be to remove some of them (i.e. I found 4 or 5 different files for the white "Star" of profile...) and normalize size, thickness or colors (for colors it's easier because we can use color palet...)

Issue #2721, opened 2020-06-16
2.6.7-dev BG deltas consistency


chrisbev:
Top right BG deltas, for me the delta includes '+' sign for positive numbers, the 15m and 40m deltas do not. 15m and 40m include a ':' after but the delta does not. 
Samsung A320.

Reporting bugs
--------------
- **Note the precise time the problem occurred** and describe the circumstances and steps that caused
  the problem
- Note the Build version (found in the About dialog in the app, when pressing the three dots in the
  upper-right corner).
- Obtain the app's log files, which can be found on the phone in
  _/storage/emulated/0/Android/data/info.nightscout.androidaps/_
  See https://github.com/MilosKozak/AndroidAPS/wiki/Accessing-logfiles
- Open an issue at https://github.com/MilosKozak/AndroidAPS/issues/new

chrisbev:
Just as another silly thought if say the time prefixed the delta triangles then the triangles could be in a verticle line?
         A: +0.2
15m A: +0.3
40m A: +0.4
Need fixed spacing?
Or 05m: +0.2?

MilosKozak:
screenshot?

chrisbev:
![Screenshot_20200617-122342_AndroidAPS](https://user-images.githubusercontent.com/43484095/84893482-37be7380-b097-11ea-951b-dcdc91ba215d.jpg)
![Screenshot_20200617-130231_AndroidAPS](https://user-images.githubusercontent.com/43484095/84895924-fd56d580-b09a-11ea-9a3c-fc56892bd725.jpg)

Philoul:
See proposed PR and result
![BGDeltasEnhancement](https://user-images.githubusercontent.com/52934600/85759828-8f9d5000-b711-11ea-8a59-70edc3d3694b.jpg)

chrisbev:
Looks excellent cheers...

Philoul:
PR updated to align values to the right too...
See last screenshot here: https://github.com/MilosKozak/AndroidAPS/pull/2775

chrisbev:
Looks good thanks, Samsung A3 2017 A320 Android 8.0 portrait/landscape. Samsung GT-i8190 A7.1.2. 799d4205d Not sure if relevant or something else, see screenshots A3 is loop phone and GT-i8190 is using NS as BG source, see 40m on GT it is zero with no sign but A3 has +0.2. 
[...]

PR #2775, opened 2020-06-25
BG Deltas consistency enhancement

Philoul:
- Delta after minutes and label right aligned in layout (label hard coded in layout)
- Signed delta avg values (note "0" value remains unsigned as before...)

See https://github.com/MilosKozak/AndroidAPS/issues/2721

and result on my loop phone (also verified on my low res dev phone) :
BGDeltasEnhancement


MilosKozak:
should not be values aligned to the right ?

Philoul:
You're right, we can also align the values ​​to the right (but I think it's better to keep the labels aligned to the right too)

2nd proposition with both labels and values aligned to the right

with screenshot
Screenshot_20200627-014112_AndroidAPS

Issue #2796, opened 2020-07-04
2.6.7-beta2 Safety fingerprint scan can always be circumvented


ArthurusDent:
The safety fingerprint scan can always be circumvented.

#### Time: 
Always when triggered.

#### Steps to reproduce: 
1. Go to the right upper corner under Preferences and then General > Advanced > Protection. Enable protection with biometrics (fingerprint scanner) for Settings, Application, and Bolus.

2. Whenever you are being asked to provide your fingerprint you can circumvent the check by tapping in the AndroidAPS window behind the safety fingerprint dialogue or by pressing the back button (haven't tested with gestures because I don't use them). 

Result: Any treatment that's being put in will be executed and shows up under Treatments. 

Opening the app:
![Screenshot_20200704-174013](https://user-images.githubusercontent.com/6848629/86516307-cebd5680-be1f-11ea-9545-8492c56c22df.jpg)

Entering carbs:
![Screenshot_20200704-174020](https://user-images.githubusercontent.com/6848629/86516308-cfee8380-be1f-11ea-99bb-daa57e5d0592.jpg)

Entering a bolus:
![Screenshot_20200704-174028](https://user-images.githubusercontent.com/6848629/86516310-d1b84700-be1f-11ea-9985-32ec2d7c2bbf.jpg)

#### Build number: 
157d8f7f5

#### Reproducability:
Oneplus 3T 
Virtual Pump (Medtronic 754)
Dexcom G6 @ xDrip+

Google Pixel 3a
Dana RS
Dexcom G6 @ xDrip+

#### Log file:
[AndroidAPS.log](https://github.com/MilosKozak/AndroidAPS/files/4873115/AndroidAPS.log)

AdrianLxM:
@MilosKozak my educated guess: `BiometricConstants.ERROR_USER_CANCELED` and `BiometricConstants.ERROR_CANCELED` will lead to a fallback to "Master password" and if that is not set, it will succeed.

MilosKozak:
do you have master password set?

MilosKozak:
master password is a fallback when you are not able authenticate by biometric

AdrianLxM:
@ArthurusDent @MilosKozak Fix: https://github.com/MilosKozak/AndroidAPS/pull/2799

PR #2799, opened 2020-07-04
Warn user of insecure fallback on missing master password

AdrianLxM:
Issue:
If no master password is set, the fallback when cancelling a biometric challange (for a bolus e.g.) always succeeds.
https://github.com/MilosKozak/AndroidAPS/issues/2796

Solution:
1) Warn the user if no master password is set when activating biometric protection.
2) Warn the user of unsecure fallback if the settings are erased.


MilosKozak:
👍

PR #2817, opened 2020-07-12
Medtronic modules

andyrozman:
- Created Rileylink and Medtronic module
- moved required files
- refactored some classes that were not correctly implemented for multiple module arhitecture 
- hardcoded doubleBolusDebug to false (settings ignored) so that we don't produce exception when using gson


app/src/main/java/info/nightscout/androidaps/activities/MyPreferenceFragment.kt
```diff
@@ -76,7 +76,7 @@ class MyPreferenceFragment : PreferenceFragmentCompat(), OnSharedPreferenceChang
     @Inject lateinit var insulinOrefFreePeakPlugin: InsulinOrefFreePeakPlugin
     @Inject lateinit var loopPlugin: LoopPlugin
     @Inject lateinit var localInsightPlugin: LocalInsightPlugin
-    @Inject lateinit var medtronicPumpPlugin: MedtronicPumpPlugin
+    @Inject lateinit var medtronicPumpPlugin: info.nightscout.androidaps.plugins.pump.medtronic.MedtronicPumpPlugin
```
MilosKozak, line 79: look at the diff and fix lines like this

MilosKozak:
how much is it tested?

app/src/main/java/info/nightscout/androidaps/activities/MyPreferenceFragment.kt
andyrozman: Should be fixed now.

PR #2832, opened 2020-07-20
Fix to issue#1617

tomvanacker85:
Added units to IC, ISF and Target in local profile


MilosKozak:
can you create this patch based on current dev?

Issue #1617, opened 2019-02-02
Units missing when entering a local basal profile


UlricusR:
Build version: c7ab2ae47-2019.01.31-20:56

Units are missing when using the local profile. This may lead to incorrect entries by users and this is a potential risk. Suggestion is to add units at least in the heading of each section, e.g.:
- DIA [h]
- IC [g/U]

![img_2415](https://user-images.githubusercontent.com/44271004/52168223-ba0af980-2727-11e9-98c5-b73a342293a8.JPG)

PR #2845, opened 2020-07-22
Added units to IC, ISF and Target in local profile (Issue #1617)

tomvanacker85:


app/src/main/res/values/strings.xml
```diff
@@ -351,7 +351,7 @@
     <string name="button3">Button 3</string>
     <string name="units_colon">Units:</string>
     <string name="units">Units</string>
-    <string name="dia">DIA</string>
+    <string name="dia">Duration of Insulin Action</string>
```
MilosKozak, line 354: did you check there is enough space on all place where used?

MilosKozak:
@tomvanacker85

PR #2887, opened 2020-08-14
Replace Default Dia by profile Dia in extendedBolus's Iob calculation

Philoul:
Currently, in `ExtendedBolus.java` file, `dia_ago` is calculated with `contants.defautDIA` (5 hours) and not with profile dia
If profile dia is longer than default dia, each iob contrib is calculated with profile dia in `InsulinOrefBasePlugin` but contribs between Profile dia and default dia are not integrated in extendedBolus's iob calculation...

=> I propose this PR to fix that

### DIFF
--- a/core/src/main/java/info/nightscout/androidaps/db/ExtendedBolus.java
+++ b/core/src/main/java/info/nightscout/androidaps/db/ExtendedBolus.java
@@ -65,6 +65,9 @@ public class ExtendedBolus implements Interval, DataPointWithLabelInterface {
     @DatabaseField
     public int insulinInterfaceID = InsulinInterface.OREF_RAPID_ACTING;
 
+    @DatabaseField
+    public double dia = Constants.defaultDIA;
+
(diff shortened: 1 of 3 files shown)


core/src/main/java/info/nightscout/androidaps/db/ExtendedBolus.java
```diff
@@ -64,8 +64,6 @@
 
     @DatabaseField
     public int insulinInterfaceID = InsulinInterface.OREF_RAPID_ACTING;
-    @DatabaseField
-    public double dia = Constants.defaultDIA;
```
MilosKozak, line 68: well changing database fields has lof of sideffects
it should be done along with increasing buildnumber and leads to braking backward compatibility

MilosKozak: i agree with changes but keep this field

PR #2902, opened 2020-08-18
Fix too long number in "adj. req. rate: 1.2000000000000000002"

GermanBluefox:
image


philou78:
Hello,

This issue also occurs with all 'rate + "U/hr"' strings.
Is it possible to change 'rate' with 'round (rate, 2)' also for these strings?

Isn't changing "rate" with "rate.toFixed(2)" easier?

PR #2908, opened 2020-08-20
Store LastRun SMB data in their dedicated fields instead of overwriting TBR fields

bartsopers:
In LoopPlugin, SMB data would be set in the TBR fields instead of the SMB fields. This PR fixes that and stores the SMB data in their dedicated fields.

In the Loop tab in AAPS, this would cause the TBR fields to contain SMB information, and the SMB fields to always be empty. See below screenshot for an example.

Screenshot_20200820-110241

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
@@ -537,11 +537,11 @@ public class LoopPlugin extends PluginBase implements LoopInterface {
                                 applySMBRequest(resultAfterConstraints, new Callback() {
                                     @Override
                                     public void run() {
-                                        //Callback is only called if a bolus was acutally requested
+                                        // Callback is only called if a bolus was actually requested
                                         if (result.enacted || result.success) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
```diff
@@ -539,13 +539,13 @@ public void run() {
...
-                                            lastRun.setLastTBRRequest(lastRun.getLastAPSRun());
-                                            lastRun.setLastTBREnact(DateUtil.now());
+                                            lastRun.setSmbSetByPump(result);
+                                            lastRun.setLastSMBRequest(lastRun.getLastAPSRun());
+                                            lastRun.setLastSMBEnact(DateUtil.now());
                                         } else {
                                             new Thread(() -> {
                                                 SystemClock.sleep(1000);
-                                                invoke("tempBasalFallback", allowNotification, true);
+                                                invoke("tbrBasalFallback", allowNotification, false);
```
bartsopers, line 548: I'm not 100% sure if the last parameter should be changed. Please confirm.

PR #2936, opened 2020-08-29
Add option to gradually increase alert volume

bartsopers:
Currently there is no way to set the volume for alerts coming from AndroidAPS. The volume is always set to the maximum value, which can be quite inconvenient. This PR adds the option to gradually increase the alert volume.


smuething:
I helped out Bart by fixing a formatting problem (he is away from his computer at the moment)

PR #2947, opened 2020-09-03
Fix 2 bugs in ValidatingEditTextPreference

bartsopers:
This PR fixes 2 bugs in ValidatingEditTextPreference:
- Fixes a bug where validations would not be applied when editing a setting more than once without leaving the corresponding settings page
- Fixes a bug where it was possible to set an invalid value


bartsopers:
These changes are also in PR #2945.

PR #2953, opened 2020-09-09
Make need for additional PIN at end of security code more obvious

dlvoy:
Make need for additional PIN at end of security code (after OTP token) more obvious:

* change hint for code checking edit field to indicate need for PIN
* add additional text with help how code is constructed from OTP token and PIN
* change PIN preference label - add that it is **mandatory**
* add _followed by PIN_ to text sent via SMS
* change order of config fields in SMS communicator plugin preferences, to have PIN configured first before authenticator


osodebailar:
Now the hints are clear enough to tell the user to generate the password.

PR #2960, opened 2020-09-10
Change obj. 9 + remove skipping objectives

xJoe:
- MA is eliminated in 2.7
- AMA is standard algo
- Basically obj. 9 is no longer needed but still there
- Tried to write an explenation (see also docs 2.7 https://github.com/openaps/AndroidAPSdocs/blob/2.7_docs/docs/EN/Usage/Objectives.rst#objective-9-try-additional-features-for-daytime-use-and-gain-confidence-in-your-closed-loop-system)
- New string "objectives_ama_gate" (Don't know if it is that easy to implement but looked how it works for obj. 10 / smb)


xJoe:
Skipping objectives is not available for the time being

PR #2961, opened 2020-09-10
Removed skipping objectives

xJoe:
Skipping objectives is not available for the time being


xJoe:
Wrong PR destination. Sorry.

